PR #34, opened 2019-08-23
chore: 🤖 plugin

NetanelBasal:
TODO: README FILE


projects/ngneat/transloco/src/lib/helpers.ts
```diff
@@ -54,3 +54,7 @@ export function mergeDeep(target: Object, ...sources: Object[]) {
 
   return mergeDeep(target, ...sources);
 }
+
+export function isNotBrowser() {
```
DmitryEfimenko: I'd reverse this to get rid of the negative: isBrowser

projects/ngneat/transloco/src/lib/transloco.service.ts
```diff
@@ -223,6 +223,36 @@ export class TranslocoService {
...
   }
 
+  /**
+   * Returns the language code name from the browser, e.g. "de"
+   *
+   * @example
+   *
+   * service.getBrowserLang()
+   */
+  getBrowserLang() {
```
DmitryEfimenko: It's great that `TranslocoService` provides a helper method: `getBrowserLang`, but it should also provide a method similar to the one in ngx-translate: [getBrowserCultureLang](https://github.com/ngx-translate/core/blob/master/projects/ngx-translate/core/src/lib/translate.service.ts#L514) since the app might use the long version of the locale (ex: en-US)

projects/ngneat/transloco-persist-lang/src/lib/persist-lang.service.ts
```diff
@@ -0,0 +1,69 @@
...
+    this.service.langChanges$.pipe(skip(1)).subscribe(lang => {
+      this.save(lang);
+    });
+  }
+
+  private setActiveLang() {
+    const cached = this.getStorage().getItem(this.storageKey);
+    const browserLang = this.service.getBrowserLang();
+    const defaultLang = this.service.config.defaultLang;
+    const currentLang = cached || browserLang || defaultLang;
```
DmitryEfimenko: There are multiple issues with this method.

I understand that this method will only be called if the strategy is `auto`, which is great. However, I think it's implementation will result in trouble.

The first issue is that the library makes a huge assumption on behalf of the developer that the lang format is short - something like *en* rather than *en-US*. One application I worked on used long formats. If the lib makes this assumption, the `currentLang` will end up in the cache as `en`, which will most likely result in an HTTP request for en.json file instead of `en-US.json`, which will return 404.

The second issue is the order of determining the `currengLang`. For example, if a developer provides a `defaultLang` configuration value, it might come as a surprise for the developer to see that this value is not being picked up if there is nothing in the cache. The `browserLang` would get picked up leaving the `defaultLang` kind of useless.

If you'd like to leave the `auto` strategy in place, at least change the order to `const currentLang = cached || defaultLang || browserLang;`
However, I'm wondering if it'd be better to leave it up to the consumer application to implement proper logic of figuring out the current language.
Another way would be to add a new property into the TRANSLOCO_PERSIST_LANG_CONFIG: something like:
```
getLang: (cachedLang: string, defaultLang: string, translocoService: TranslocoService) => string;
```
With this, the developer will be able to specify the logic for determining the language in the consumer app.

projects/ngneat/transloco/src/lib/types.ts
```diff
@@ -18,3 +18,4 @@ export type FailedEvent = {
 export type TranslocoEvents = LoadedEvent | FailedEvent;
 export type Translation = HashMap<any>;
 export type TranslationCb<T> = (translation: T, params?: HashMap) => string;
+export type PersisStorage = Pick<Storage, 'getItem' | 'setItem' | 'removeItem'>;
```
DmitryEfimenko: typo: PersistStorage instead of PersisStorage

PR #65, opened 2019-09-05
fix(core): lack of unsubscriptions

DerSizeS:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[x] Code style update (formatting, local variables)
[x] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

In some cases, there is no unsubscription, which makes memory leaks possible.

## What is the new behavior?

Always unsubscribe when necessary.

```
[ ] Yes
[x] No
```

## Other information
OnDestroy lifecycle hook works in [services](https://angular.io/api/core/OnDestroy) and [modules](https://github.com/angular/angular/blob/master/packages/core/src/view/ng_module.ts#L222).
Even if the service is provided in root, it can be destroyed when the platform is [destroyed manually](https://angular.io/api/core/PlatformRef#destroy).

### DIFF
--- a/projects/ngneat/transloco-persist-lang/src/lib/persist-lang.service.ts
+++ b/projects/ngneat/transloco-persist-lang/src/lib/persist-lang.service.ts
@@ -6,16 +6,18 @@ import {
   PersistStorage,
   TranslocoService
 } from '@ngneat/transloco';
+import { Subscription } from 'rxjs';
 import { skip } from 'rxjs/operators';
 import {
   PersistLangConfig,
   TRANSLOCO_PERSIST_LANG_CONFIG,
   TRANSLOCO_PERSIST_LANG_STORAGE
 } from './persist-lang.config';
-import { Inject, Injectable } from '@angular/core';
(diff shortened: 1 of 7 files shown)


NetanelBasal:
Thanks for the effort, but this is by design. All our services our wide-app singletons so there is no need to unsubscribe.

DerSizeS:
> Thanks for the effort, but this is by design. All our services our wide-app singletons so there is no need to unsubscribe.

Yes, your services are all singleton and are provided at the **application level**. But this does not mean that you do not need to unsubscribe. The application can be destroyed not only by closing the browser window, but also [manually](https://angular.io/api/core/NgModuleRef#destroy) ([showcase](https://stackblitz.com/edit/angular-wcdrrg?embed=1&file=src/main.ts)).

I don’t think that unsubscribing is bad for your code, on the contrary, it’s rational.

If I have not convinced you, feel free to close the PR.

NetanelBasal:
Although it's a rare case, I will merge your code soon. It's also valid for applications that work with HMR.

PR #116, opened 2019-10-04
fix/add-key-prefix-when-translating-scope

shaharkazaz:


projects/ngneat/transloco/src/lib/transloco.service.ts
```diff
@@ -163,11 +163,21 @@ export class TranslocoService implements OnDestroy {
    * translate('scope.someKey', { }, 'en')
    */
   translate<T = any>(key: TranslateParams, params: HashMap = {}, lang = this.getActiveLang()): T {
-    const resolveLang = this._isLangScoped(lang) ? this.getActiveLang() : lang;
+    let resolveLang = lang;
+    let scope;
+    if (this._isLangScoped(lang)) {
+      const langFromScope = getLangFromScope(lang);
+      const hasLang = this._isLang(langFromScope);
```
itayod: I think we better throw an error if the language is not valid.

projects/ngneat/transloco/src/lib/transloco.service.ts
```diff
@@ -163,11 +163,21 @@ export class TranslocoService implements OnDestroy {
...
+      resolveLang = hasLang ? langFromScope : this.getActiveLang();
+      scope = this.getMappedScope(hasLang ? getScopeFromLang(lang) : lang);
+    }
+
     if (Array.isArray(key)) {
-      return key.map(k => this.translate(k, params, resolveLang)) as any;
+      return key.map(k => this.translate(scope ? `${scope}.${k}` : k, params, resolveLang)) as any;
     }
 
+    key = scope ? `${scope}.${key}` : key;
```
itayod: DRY

PR #44, opened 2019-08-27
feat(directive): add ability to read nested property

fxck:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/ngneat/transloco/blob/master/CONTRIBUTING.md#commit
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?
Adds ability to limit a translation to a nested property so you don't have to write the full path in each translation.

Given this:

```typescript
{
  common: {
    foo: 'Foo',
    bar: 'Bar'
  },
  dashboard: {
    title: 'Title',
    desc: 'Desc'
  }
}
```

you can do

```html
<ng-container *transloco="let t; limit: 'dashboard'">
  <h1>{{ t.title }}</h1>
  <p>{{ t.desc }}</p>
</ng-container>
```

```
[ ] Bugfix
[x] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] Other... Please describe:
```

## What is the current behavior?

Issue Number: #41

## Does this PR introduce a breaking change?

```
[ ] Yes
[x] No
```

## Other information

1. not sure about the naming
2. should it support deeply nested properies `dashboard.some.deeply.nested.obj`? 

@NetanelBasal

PR #219, opened 2020-01-15
feat(directive): add ability to get full nested object

nimesko:
## PR Checklist

- [X] The commit message follows our guidelines: https://github.com/ngneat/transloco/blob/master/CONTRIBUTING.md#commit
- [X] Tests for the changes have been added (for bug fixes / features)
- [X] Docs have been added / updated (for bug fixes / features)

## PR Type

This PR is similar to this one (https://github.com/ngneat/transloco/pull/44) but the difference is that we can retrieve the full nested object.

You can imagine: 
> "Why I need the reference of object?"

And I answer:
> "Because we can use smart/dumb components pattern and need the parent to control the translations of your childs, making then more unbound"

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: [221](https://github.com/ngneat/transloco/issues/221)

## What is the new behavior?

As described in the issue, the objective is to support the return of new objects by adding in the value `true` in the second parameter or, if you need to pass information in the second parameter (parameterized translation), put in the third parameter, the value `true`.

```
[ ] Yes
[X] No
```

## Other information

1. maybe we can pass more properties and add a way that we pass an object (instead multiple parameters) to control the behavior of translation

### DIFF
--- a/projects/ngneat/transloco/src/lib/tests/directive/structural.spec.ts
+++ b/projects/ngneat/transloco/src/lib/tests/directive/structural.spec.ts
@@ -15,6 +15,7 @@ describe('Structural directive', () => {
            <span>{{t('fromList')}}</span>
            <p>{{t('a.b.c')}}</p>
            <h2>{{t('a.b.c', {fromList: "value"}) }}</h2>
+           <a>{{t('nested', true) | json }}</a>
         </section>
      `);
     runLoader();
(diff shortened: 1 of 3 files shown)


itayod:
Hi @nimesko 
Thanks for the pr (next time I suggest you would open an issue first :) ).
for your use case did you consider using `selectTranslationObject()` method from the service?

nimesko:
Hi @itayod,
Sorry for not creating the issue, can I create the issue there and continue the debate there?

About the `selectTranslationObject`: Yes, I considered, the point is that it returns an observable and, if we were to use the same structural directive, we would have to pass `pipe async` in the template being different if the implementation requires both forms (with the return of the translated object / with the translated term)

Issue #317, opened 2020-07-28
Add schematic for transloco-keys-manager does not observe configured package manager.


napei:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ x ] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request
[ ] Other... Please describe:
</code></pre>

## Current behavior

This is an issue with the schematic for [ngneat/transloco-keys-manager](https://github.com/ngneat/transloco-keys-manager), however the schematic is in this repo so I'll add the issue here.

The schematic explicitly uses npm to install packages, even though it's explicitly set in `angular.json` that I use `yarn`, and a `yarn.lock` is present. This is kind of annoying as I need to then remove the `package-lock` file and then reinstall with `yarn` again.

[https://github.com/ngneat/transloco/blob/c5281191bab02b516df9d4cc2ab4711d25836678/schematics/src/keys-manager/index.ts#L11](https://github.com/ngneat/transloco/blob/c5281191bab02b516df9d4cc2ab4711d25836678/schematics/src/keys-manager/index.ts#L11)

## Expected behavior

I expect the `ng add` schematic to use whatever angular is configured to use which I believe is available under `NodePackageInstallTask()` in `@angular-devkit/schematics/tasks`

## Minimal reproduction of the problem with instructions

- Generate new angular application

- Set package manager to yarn with `ng config cli.packageManager yarn` or manually by editing `angular.json` and adding the following:

 ```json
  "cli": {
     "packageManager": "yarn"
   }
```

- Install `transloco-keys-manager` using `ng add @ngneat/transloco-keys-manager`

## What is the motivation / use case for changing the behavior?

Annoying that it doesn't follow the angular configuration.

## Environment

<pre><code>
Angular version: 10.0.5

Browser:
N/A

For Tooling issues:
- Node version: 14.6.0  
- Platform: Windows  

Others:

</code></pre>

itayod:
@napei would you like to create a PR for that?

napei:
@itayod Just gave it a go but couldn't get anything to build properly on my machine. Also I'm on windows and `cp` doesn't exist so the copy npm scripts don't work. Was also getting a `TypeError: this.typeChecker.getTypeArguments is not a function` from `tsickle`. It's probably more efficient for you guys to add it. Sorry about that 😢 

Basically I got as far as adding this in `index.ts` for `keys-manager`

[...]

PR #322, opened 2020-07-30
fix: 🐛 update schematic for keys-manager

napei:
Changes the schematic for keys-manager to use angular-defined package managers. Versions of the package to install are stored as constants. Fixes: #317 

## PR Checklist

## PR Type

```
[ x ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently `ng g @ngneat/transloco:keys-manager` will forcibly use npm to install packages even though the user may be using a different package manager.

Issue Number: N/A

## What is the new behavior?

The schematic will use built in functions to the schematics library to install the packages. Further to that `ngx-build-plus` was being installed regardless of if the user selected `CLI` in the initial options. This was refactored into the same conditional used for `Webpack` and `Both`. The only notable change is that this requires the package version of `@ngneat/transloco-keys-manager` and `ngx-build-plus` to be kept up to date in `schematics/src/schematics.consts.ts`

- [ ] Yes
- [x] No

## Other information

Some other minor refactoring was performed by adding types to some function parameters that were missing them in that schematic.

This schematic was tested on my local machine on an Angular 10 project, and worked fine.

### DIFF
--- a/schematics/src/keys-manager/index.ts
+++ b/schematics/src/keys-manager/index.ts
@@ -1,17 +1,20 @@
-import { Rule, Tree, SchematicsException } from '@angular-devkit/schematics';
+import { Rule, Tree, SchematicsException, SchematicContext } from '@angular-devkit/schematics';
 import { TranslocoConfig } from '@ngneat/transloco-utils';
-import { execSync } from 'child_process';
 import { addScriptToPackageJson } from '../utils/package';
 import { getWorkspace, setWorkspace } from '../utils/projects';
 import { getConfig, updateConfig } from '../utils/transloco';
 import { SchemaOptions } from './schema';
 
-function installKeysManager() {
-  console.log('Installing packages for tooling...');
-  execSync('npm install --save-dev @ngneat/transloco-keys-manager ngx-build-plus');
+import { addPackageJsonDependency, NodeDependencyType } from '@schematics/angular/utility/dependencies';
+import { NodePackageInstallTask } from '@angular-devkit/schematics/tasks';
+import { KEYS_MANAGER_VERSION, NGX_BUILD_PLUS_VERSION } from '../schematics.consts';
+
+function installDependencies(context: SchematicContext) {
+  context.logger.log('info', `🔍 Installing packages...`);
+  context.addTask(new NodePackageInstallTask());
 }
 
-export function updateAngularJson(host: Tree, options) {
+export function updateAngularJson(host: Tree, options: SchemaOptions) {
   const angularJson = getWorkspace(host);
   if (angularJson) {
     const project = angularJson.projects[options.project || angularJson.defaultProject];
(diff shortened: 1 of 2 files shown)


schematics/src/schematics.consts.ts
```diff
@@ -2,3 +2,5 @@ export const PROJECT_NAME = 'transloco';
 export const NPM_SCOPE = '@ngneat';
 export const LIB_NAME = '@ngneat/transloco';
 export const CONFIG_FILE = 'transloco.config.js';
+export const KEYS_MANAGER_VERSION = '2.2.1';
```
itayod, line 5: what if we update the keys manager version? I wouldn't like updating it over and over...
p.s. moving the script to the keys-manager lib will fix it but I am not sure we like it... (maybe we will do it in the future but currently this is not the solution I am looking for...).

schematics/src/keys-manager/index.ts
```diff
@@ -1,17 +1,20 @@
...
-function installKeysManager() {
-  console.log('Installing packages for tooling...');
-  execSync('npm install --save-dev @ngneat/transloco-keys-manager ngx-build-plus');
+import { addPackageJsonDependency, NodeDependencyType } from '@schematics/angular/utility/dependencies';
+import { NodePackageInstallTask } from '@angular-devkit/schematics/tasks';
+import { KEYS_MANAGER_VERSION, NGX_BUILD_PLUS_VERSION } from '../schematics.consts';
+
+function installDependencies(context: SchematicContext) {
+  context.logger.log('info', `🔍 Installing packages...`);
+  context.addTask(new NodePackageInstallTask());
```
itayod, line 14: maybe a better solution would be figuring out if the project is using yarn or npm and simply run the correct command?

schematics/src/schematics.consts.ts
napei: If you update keys-manager then you update the version that the schematic installs. It's annoying however it's better to do it this way, than to forcibly run a command like `npm install`

schematics/src/keys-manager/index.ts
napei: That's probably too much work for a schematic? The tooling for adding packages and executing the package manager for the project already exist as provided from the angular team.

schematics/src/schematics.consts.ts
napei: Further to that, packages like angular itself hard code the version, or at least they hard code the assumption to use the current version of the cli, as all angular packages are supposed to have the same semver. So instead of doing that, in this case, you would need to keep track of which version to explicitly install. 

Technically you can use `latest` as the version, however I don't think that's good practice to have `latest` in the package.json

schematics/src/keys-manager/index.ts
napei: > maybe a better solution would be figuring out if the project is using yarn or npm and simply run the correct command?

And assuming the package manager is not going to be 100% accurate all the time.

schematics/src/schematics.consts.ts
napei: > p.s. moving the script to the keys-manager lib will fix it but I am not sure we like it... (maybe we will do it in the future but currently this is not the solution I am looking for...).

Moving the script to the keys-manager lib actually makes more sense than keeping it in this lib, as you then don't need to update the version in this repo, and semantically you are adding keys-manager to the project, which is a dev dependency, through `ng add` not `ng g` which is a little confusing anyway.

schematics/src/keys-manager/index.ts
itayod: The reason we didn't use this build-in tool is that we needed to specify the version instead of installing the latest version, I would look how angular decide to use yarn or npm and use the same method.

napei: > The reason we didn't use this build-in tool is that we needed to specify the version instead of installing the latest version and as far as I know, Angular's API does not support that, 

The angular API installs a specified version.

itayod: yap, so that's why it isn't good enough for us :)

napei: If you put this script into the keys-manager repo, this problem will go away.

>  we needed to specify the version instead of installing the latest version and as far as I know, Angular's API does not support that


I think you meant you "needed to install the latest version instead of specifying the version.."

schematics/src/schematics.consts.ts
itayod: > > p.s. moving the script to the keys-manager lib will fix it but I am not sure we like it... (maybe we will do it in the future but currently this is not the solution I am looking for...).
> 
> Moving the script to the keys-manager lib actually makes more sense than keeping it in this lib, as you then don't need to update the version in this repo, and semantically you are adding keys-manager to the project, which is a dev dependency, through `ng add @ngneat/transloco-keys-manager` instead of `ng g @ngneat/transloco:keys-manager` which is generating something rather than adding something

Sure, but as I said we currently do not want to do it... also, keep in mind it is not an Angular project so I am not sure if it even possible right now... anyway I am looking for a simple solution that detect if the user is using `yarn` as a package manager (I think you can get that from the `angular.json`), and run `yarn` or `npm` accordingly.

Issue #323, opened 2020-07-30
Specify start and end parameter value instead of having to overwrite the transpiler


erikmom:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[x] Feature request
[ ] Documentation issue or request
[ ] Support request
[ ] Other... Please describe:
</code></pre>

## Current behavior

In our application we use a mixture of frameworks (Java, Angular) and all of these frameworks read from the same message bundle. Unfortunately these frameworks read the message bundle in a different way regarding the parameters.

An example value in our message bundle : "Failed to delete {0}"

Note the single curly brackets, while transloco expects double curly brackets by default.

To workaround this we have made a custom transpiler, but the only thing that is changed in comparison with the default transpiler is replacing the double curly brackets with single curly brackets.

## Expected behavior

It would be nice if we could just specify a start and end parameter value, instead of overwriting the default transpiler with a custom transpiler.

## What is the motivation / use case for changing the behavior?

Less code to maintain and less likely that things break while upgrading to newer versions of transloco, e.g. if something changes in the default transpiler.

shaharkazaz:
@erikmom Would you like to create a PR for it? 😄

shaharkazaz:
Thanks @Waterstraal for the PR! realease in `transloco v2.19.0` & `transloco-messageformat v1.3.0`

PR #330, opened 2020-08-14
feat: 🎸 Add interpolation to configuration

Waterstraal:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

You cannot specify your own regular expression for matching parameters. 

Issue Number: #323 

## What is the new behavior?

A new transpiler has been added that adds the possibility to specify your own regular expression for matching parameters. 

```
[ ] Yes
[x] No
```

### DIFF
--- a/docs/docs/transpiler.mdx
+++ b/docs/docs/transpiler.mdx
@@ -6,6 +6,38 @@ description: The Transpiler | Transloco Angular i18n
 The transpiler is responsible for resolving the given value.  
 For example, when given `Hello {{ key }}` the default transpiler will replace the dynamic variable `key` based on the given params, or in some cases, within the [translation object itself](./additional-functionality#reference-other-keys).
 
+## DefaultTranspiler
+The default transpiler can be configured with a custom interpolation regular expression or start and end markings to match message parameters.
+
+When passing in a regular expression, the `DefaultTranspiler` expects one capturing group and the global flag to be enabled.
+
+To configure the `DefaultTranspiler` you must provide a Transloco config with the interpolation property set. For example:
+
+```ts title="transloco-root.module.ts"
+import { TRANSLOCO_CONFIG, translocoConfig } from '@ngneat/transloco';
+
+@NgModule({
+  ...
+  providers: [
+    {
+      provide: TRANSLOCO_CONFIG,
+      useValue: translocoConfig({
+        interpolation: /<<<(.*?)>>>/g // or ['<<<', '>>>']
+      })
+    }
+  ]
+})
+export class TranslocoRootModule {}
+```
+
+Your message bundle must now specify parameters with triple angle brackets:
+```json title="en.json"
+{
+  "hello": "Hello <<<value>>>"
+}
+```
+
+
 ## Functional Transpiler
 
 In addition to the default transpiler, Transloco also exposes the `FunctionalTranspiler` which allows you to implement function calls into your translation values. This way you can leverage Angular's DI power and make your translations even more flexible.
@@ -115,33 +147,5 @@ export class CustomTranspiler implements TranslocoTranspiler {
 export const customTranspiler = {
   provide: TRANSLOCO_TRANSPILER,
   useClass: CustomTranspiler
-}
-```
-
-## RegExTranspiler
-
-In addition to the default transpiler, Transloco also exposes the `RegExTranspiler` which allows you to specify your own regular expression to match message parameters.
-
-The `RegExTranspiler` expects one capturing group for the parameters.
-
-To enable this transpiler, add the following provider in your `TranslocoRootModule` (provided with your own regex):
-
-```ts title="transloco-root.module.ts"
-import { RegExTranspiler, TRANSLOCO_TRANSPILER } from '@ngneat/transloco';
-
-@NgModule({
-  ...
-  providers: [{
-    provide: TRANSLOCO_TRANSPILER,
-    useFactory: () => new RegExTranspiler(/<<<(.*?)>>>/g)
-  }]
-})
-export class TranslocoRootModule {}
-```
-
-Your message bundle must now specify parameters with triple angle brackets:
-```json title="en.json"
-{
-  "hello": "Hello <<<value>>>"
 }
 ```
--- a/docs/docs/transpiler.mdx
+++ b/docs/docs/transpiler.mdx
@@ -117,3 +117,31 @@ export const customTranspiler = {
   useClass: CustomTranspiler
 }
 ```
+
+## RegExTranspiler
+
+In addition to the default transpiler, Transloco also exposes the `RegExTranspiler` which allows you to specify your own regular expression to match message parameters.
+
+The `RegExTranspiler` expects one capturing group for the parameters.
+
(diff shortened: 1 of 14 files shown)


shaharkazaz:
@Waterstraal Thanks for the PR!

I don't think this should be an actual transpiler, I believe that a property in the config is more fitting here.
Can you replace the transpiler with the `interpolation` config property and consume it in the `defaultTranspiler`? 

It should support the following values:
* A Regex
* An array that indicates the start and end markings ([start, end]: [string, string]).

A couple of things to keep in mind:
* Validate that the input.
* If a regex is provided we need to make sure the global flag is present.

Waterstraal:
> @Waterstraal Thanks for the PR!
> 
> I don't think this should be an actual transpiler, I believe that a property in the config is more fitting here.
> Can you replace the transpiler with the `interpolation` config property and consume it in the `defaultTranspiler`?
> 
> It should support the following values:
> 
> * A Regex
> * An array that indicates the start and end markings ([start, end]: [string, string]).
> 
> A couple of things to keep in mind:
> 
> * Validate that the input.
> * If a Regex is provided we need to make sure the global flag is present.

Good suggestion. I pushed a new commit that adds `interpolation` to Transloco config

docs/docs/transpiler.mdx
```diff
@@ -6,6 +6,38 @@ description: The Transpiler | Transloco Angular i18n
 The transpiler is responsible for resolving the given value.  
 For example, when given `Hello {{ key }}` the default transpiler will replace the dynamic variable `key` based on the given params, or in some cases, within the [translation object itself](./additional-functionality#reference-other-keys).
 
+## DefaultTranspiler
```
shaharkazaz, line 9: The documentation for the new property should be part of the [config options page](https://ngneat.github.io/transloco/docs/config-options/).

projects/ngneat/transloco/src/lib/transloco.transpiler.ts
```diff
@@ -72,6 +80,19 @@ export class DefaultTranspiler implements TranslocoTranspiler {
   }
 }
 
+function getInterpolationRegExpFromConfig(userConfig?: TranslocoConfig): RegExp {
+  const interpolation = userConfig ? userConfig.interpolation : defaultConfig.interpolation;
```
shaharkazaz, line 84: Looking at the code now, I don't think we need to support Regex, don't see any reason for someone to pass a regex instead of the markings. Do you agree?

Update the default value to: `['{{', '}}']`
```
const [start, end] = userConfig?.interpolation ?? defaultConfig.interpolation;

return new RegExp(`${start}(.*?)${end}`, 'g');
```

projects/ngneat/transloco/src/lib/transloco.transpiler.ts
```diff
@@ -1,18 +1,26 @@
...
 export const TRANSLOCO_TRANSPILER = new InjectionToken('TRANSLOCO_TRANSPILER');
 
 export interface TranslocoTranspiler {
   transpile(value: any, params: HashMap, translation: HashMap): any;
+
   onLangChanged?(lang: string): void;
 }
 
 export class DefaultTranspiler implements TranslocoTranspiler {
+  protected interpolationRegExp: RegExp;
```
shaharkazaz, line 15: How about interpolationMatcheris?

projects/ngneat/transloco/src/lib/transloco.transpiler.ts
```diff
@@ -72,6 +80,19 @@ export class DefaultTranspiler implements TranslocoTranspiler {
   }
 }
 
+function getInterpolationRegExpFromConfig(userConfig?: TranslocoConfig): RegExp {
```
shaharkazaz, line 83: How about `resolveMatcher`?

projects/ngneat/transloco/src/lib/tests/transloco.transpiler.spec.ts
```diff
@@ -86,30 +108,42 @@ describe('TranslocoTranspiler', () => {
...
+        lang
+      );
       expect(parsed).toEqual('Hello with keys from lang supporting nested values!');
     });
 
     it('should translate simple string with from lang with nested params', () => {
       const lang = flatten({
-        dear: 'dear {{name}}',
-        hello: 'Hello {{dear}}'
+        dear: `dear ${interpolationMarkings[0]}name${interpolationMarkings[1]}`,
```
shaharkazaz, line 121: create a function called `wrapParam` which replaces: `${interpolationMarkings[0]}paramName${interpolationMarkings[1]}` and replace all instances with the function call.

projects/ngneat/transloco/src/lib/transloco.transpiler.ts
Waterstraal: Makes sense to me to remove the RegExp 👍  

Sadly we can't use optional chaining and nullish coalescing as this project is still on typescript 3.2

projects/ngneat/transloco/src/lib/tests/transloco.transpiler.spec.ts
```diff
@@ -63,19 +68,27 @@ describe('TranslocoTranspiler', () => {
     });
   });
 
-  function testDefaultBehaviour(parser: TranslocoTranspiler) {
+  function testDefaultBehaviour(parser: TranslocoTranspiler, interpolationMarkings: [string, string] = ['{{', '}}']) {
     it('should translate simple string from params', () => {
-      const parsed = parser.transpile('Hello {{ value }}', { value: 'World' }, {});
+      const parsed = parser.transpile(`Hello ${wrapParam('value', interpolationMarkings)}`, { value: 'World' }, {});
```
shaharkazaz, line 73: why do you need to pass the interpolationMarkings to the function? just create it in the testDefaultBehaviour scope.

Waterstraal:
@shaharkazaz I was wondering whether we should make `TranslocoConfig` a required field in `DefaultTranspiler`'s constructor that all derived transpilers have to pass it in. This prevents a potential issue with a user configuration that specifies a custom `interpolation` in combination with a custom transpiler that doesn't pass the `TranslocoConfig` in. Your thoughts?

shaharkazaz:
@Waterstraal that would make it a breaking change and we don't want that 🙂  maybe in v3 😉 
Regarding the MessageFormatTranspiler, we need to pass the config there.

Waterstraal:
> @Waterstraal that would make it a breaking change and we don't want that 🙂 maybe in v3 😉
> 
> Regarding the MessageFormatTranspiler, we need to pass the config there, nice catch.

Shall I update `MessageFormatTranspiler`?

shaharkazaz:
@Waterstraal Yes, please test it in the playground and make sure everything is working as expected.

Waterstraal:
@shaharkazaz I updated the transpiler and added some specs and udpated cypress to verify it's working. I tested it manually on the playground by adding interpolation settings and updating message bundles. It's difficult in the current setup to have multiple instances of TranslocoConfig provided in the playground I think? Thant would be ideal to test multiple scenario's.

Issue #344, opened 2020-09-17
Bug: schematics 'join' does not look for every sub-directory 


kmathy:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request
[ ] Other... Please describe:
</code></pre>

## Current behavior

At the moment, the schematics does not take every sub-directory to create the appropriate scope.
For example:
<pre><code>
--- forms
 |
 --- validations
  |
  --- en.json
 |
 --- hints
  |
  --- en.json
|
--- en.json
</code></pre>

The last `en.json` is the root en.json.
When I launch the command `ng g @ngneat/transloco:join --default-lang en --include-defaultLang` , it does not take the en.json in forms.validations and forms.hints

To have this kind of separation, I have to create subfolders with one level depth (forms.validation, forms.hints)

However, I found something (/schematics/src/join/index.ts:20)

<pre><code>
if (!fs.existsSync(dirPath)) {
  throw new SchematicsException(`Could not resolve path to dir: ${dirPath}`);
}
if (!hasFiles(dir)) return translationJson;
</code></pre>

When I remove these lines, I have the perfect behavior. In Windows, fs.existsSync says that my sub-directory is false (but it exists)
And I removed the second 'if' to be sure it will not stop until it looked for every sub-directory

## Expected behavior

The command should take these subfiles into account. At the end, I should get a compound scope for every sub-directory discovered.

## What is the motivation / use case for changing the behavior?

The motivation behind this is to get a deeper separation of translation files.
This week-end, I could take a look to make a PR to resolve this issue

## Environment

<pre><code>
Angular version: 10.0.4

Browser:
- [x] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: v10.16
- Platform Windows 10

Others:
</code></pre>

kmathy:
Closed because Fix has been merged (PR #348)

PR #348, opened 2020-09-19
Bugfix/344: schematics join

kmathy:
## PR Checklist

## PR Type

This PR is related to the issue #344. Thanks to this change, the schematic `join` will be more precise with different scopes.

Moreover, there was a mistake of how the path is parsed between different OS. Used the `normalize` function from `@angular-devkit/core` to be sure the path will be good. Should be interesting to test to a Linux or MacOS system.

Also, with this behavior, will it be interesting to use scope differently? 

<pre><code>
{
  "key1": {
    "key2": {
      "title": "test",
      "subtitle": "sub-test"
    }
  }
}
</code></pre>

 scope: 'key1.key2' --> { "title": "test", "subtitle": "sub-test" }

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The `join` will concatenate nested keys like a path (key1.key2.[...]). 

Issue Number: #344 

## What is the new behavior?

For nested keys, it will create a child node to the parent scope. See the different pictures below:
folders
final-folder
result-en
result-es

```
[x] Yes
[] No
```

## Other information

I was not able to run the tests, I'm not sure how to proceed here


schematics/src/join/index.ts
```diff
@@ -20,22 +22,17 @@ function reduceTranslations(host: Tree, dirPath: string, translationJson, lang:
   if (!fs.existsSync(dirPath)) {
     throw new SchematicsException(`Could not resolve path to dir: ${dirPath}`);
   }
-  if (!hasFiles(dir)) return translationJson;
+  if (!hasFiles(dir) && !hasSubdirs(dir)) return translationJson;
   dir.subfiles
     .filter(fileName => fileName.includes(`${lang}.json`))
     .forEach(fileName => {
-      if (translationJson[key]) {
```
itayod, line 27: this was meant to prevent overriding duplicate key... we need to keep this check

schematics/src/utils/json-utils.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+/**
+ * This function works like the 'get' as Lodash
+ *
+ * @param object Object like a json, for example.
+ * @param path Generally a string like key1.key2 or array like ['key1', 'key2']
+ * @param defaultValue if a value is not found, provide a default value. default: undefined
+ */
+export function getProp(object, path, defaultValue = undefined) {
```
itayod, line 24: I don't see that you're using this function is it redundant? if so, please remove it

itayod (changes requested):

shaharkazaz:
@kmathy will you have time to address the comments?

kmathy:
Hi,

I did not see your messages ;) I will take some time tonight to fix the comments.

schematics/src/utils/json-utils.ts
kmathy: It was there for future improvement, but I will remove it

kmathy:
Hi @shaharkazaz  and @itayod,

I made the second change requested, however, I was not able to find a way to re-insert the previous check preventing duplicate keys.
I just discovered that `JSON.parse()` already remove duplicate keys. I wanted to make the check in this function:

```javascript
export function getJsonFileContent(fileName: PathFragment, dir: DirEntry, parser = JSON.parse) {
  return parser(dir.file(fileName).content.toString('utf-8'));
}
```
But receiving a string with the toString method, I can't make the check properly before the `JSON.parse`. I also used the second argument of this function, the `reviver`, with no success.

Do you have an idea to re-insert this check, without impacting the performance?

The best idea I had was:
```javascript
const json = getJsonFileContent(fileName, dir);
const jsonKeys = Object.keys(json);
if(jsonKeys.length !== new Set(jsonKeys).size) {
    throw new SchematicsException(
      `In file ${fileName}, there is a duplicate key , please rename it and rerun the command.`
    );
}
setProp(translationJson, key, json);
```

itayod:
> Hi @shaharkazaz and @itayod,
> 
> I made the second change requested, however, I was not able to find a way to re-insert the previous check preventing duplicate keys.
> I just discovered that `JSON.parse()` already remove duplicate keys. I wanted to make the check in this function:
> 
> ```js
> export function getJsonFileContent(fileName: PathFragment, dir: DirEntry, parser = JSON.parse) {
>   return parser(dir.file(fileName).content.toString('utf-8'));
> }
> ```
> 
> But receiving a string with the toString method, I can't make the check properly before the `JSON.parse`. I also used the second argument of this function, the `reviver`, with no success.
> 
> Do you have an idea to re-insert this check, without impacting the performance?
> 
> The best idea I had was:
> 
> ```js
> const json = getJsonFileContent(fileName, dir);
> const jsonKeys = Object.keys(json);
> if(jsonKeys.length !== new Set(jsonKeys).size) {
>     throw new SchematicsException(
>       `In file ${fileName}, there is a duplicate key , please rename it and rerun the command.`
>     );
> }
> setProp(translationJson, key, json);
> ```

Sorry for the delay, maybe I miss something, but why not do the check before you `set` the key (as it was before)?

kmathy:
Hi @itayod,

I understood my mistake: I thought the check was for the JSON file parsed, I don't know why. 

The check has been added as expected ;)

itayod:
Great @kmathy, one last thing, I am not sure that we want the `setProp` function since it separates the keys by `'.'` which might mess some stuff in the current algorithm.
please revert it to what it was before, and then I think we can merge it :)

kmathy:
And Finally, everything is done ;)

itayod:
👏 👏 LGTM

PR #357, opened 2020-10-05
plugin: get the currency symbol for the currently set locale

### LINKED ISSUE #339: [Feature Request] - getCurrentCurrency
Hi. I cant find any way to get just the current currency name or symbol.
At this moment i use `{{0| translocoCurrency: 'code' }}` and use `text-indent: -24px` to remove the 0
a function that returns the currency code / symbol / name would be very helpfull.

best regards.

PS.: I solved it by creating a pipe:

import { Pipe, PipeTransform } from '@angular/core';

````javascript
@Pipe({
  name: 'stringToCurrency'
})
export class StringToCurrencyPipe implements PipeTransform {

  transform(value: string): string {
    return value.replace(/[\d\s\,\.]/g, '');
  }

}
````

Then: `{{(0 | translocoCurrency: 'code') | stringToCurrency }}`

Bit dirty and i hope for a better solution.

### PR DESCRIPTION
**Pull request by chrsi**
There is no clean way to retrieve solely the currency symbol. Therefore a pivot element was chosen to
create a formatted number. In that result the pivot element was removed to retrieve only the currency
symbol.
Depending on the locale the symbol can be on either the left or the right side of the number.

Fixes #339

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

At the moment it is not possibly to retrieve solely the currency symbol of the currently set locale. 

Issue Number: #339 

## What is the new behavior?

The TranslocoLocaleService provides a method that is able to retrieve this symbol. This is done by formatting a 'dummy' number (pivot) and removing the number part. Whats left is basically the currency symbol.

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/ngneat/transloco-locale/src/lib/transloco-locale.service.ts
+++ b/projects/ngneat/transloco-locale/src/lib/transloco-locale.service.ts
@@ -65,9 +65,9 @@ export class TranslocoLocaleService implements OnDestroy {
   /**
    * Get the currency symbol for the currently set locale.
    */
-  getCurrencySymbol() {
-    const currency = this.localeCurrencyMapping[this._locale];
-    const numberFormat = new Intl.NumberFormat(this._locale, {
+  getCurrencySymbol(locale = this.getLocale()) {
+    const currency = this.localeCurrencyMapping[locale];
+    const numberFormat = new Intl.NumberFormat(locale, {
       style: 'currency',
       currencyDisplay: 'symbol',
       currency: currency,
--- a/projects/ngneat/transloco-locale/src/lib/transloco-locale.service.ts
+++ b/projects/ngneat/transloco-locale/src/lib/transloco-locale.service.ts
@@ -63,6 +63,28 @@ export class TranslocoLocaleService implements OnDestroy {
   }
 
   /**
+   * Get the currency symbol for the currently set locale.
+   */
+  getCurrencySymbol() {
+    const currency = this.localeCurrencyMapping[this._locale];
+    const numberFormat = new Intl.NumberFormat(this._locale, {
+      style: 'currency',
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by chrsi** [comment]
I wasn't quite sure if the methods fits into the transloco locale service. Therefore i haven't updated the docs yet. But if you think it's the right place i can go ahead and to that :)

File: projects/ngneat/transloco-locale/src/lib/transloco-locale.service.ts [hunk 1]
```diff
@@ -62,6 +62,28 @@ export class TranslocoLocaleService implements OnDestroy {
     this._locale = locale;
   }
 
+  /**
+   * Get the currency symbol for the currently set locale.
+   */
+  getCurrencySymbol() {
```
itayod, line 68: LGTM, I would just add an optional parameter for a locale something like this:
```ts
getCurrencySymbol(locale = this.getLocale()) {
```

chrsi: Yeah, good idea! Changed that

chrsi:
Added docs as well.

Would you mind participating in the hacktoberfest by adding the topic or labelling this PR as `hacktoberfest-accepted`?

projects/ngneat/transloco-locale/src/lib/transloco-locale.service.ts
itayod: Thanks, good work!

itayod:
@chrsi just one last thing, please change the commit to `plugin` instead of `feat` so it won't affect the main library's version.

chrsi:
@itayod updated PR and commit message

PR #361, opened 2020-10-09
fix(tests): fix failing unit and e2e tests

chrsi:
Some tests failed due to a missing localization of expected time values.
One e2e got removed because it isn't valid anymore.

Please check whether the localized expected time values work in your timezone as well.

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Some unit tests failed under the wrong time zone.
Some e2e tests failed because of various reasons:
* legacy
* wrong time zone

## What is the new behavior?

All unit tests and e2e tests run fine.

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/ngneat/transloco-locale/src/lib/tests/pipes/transloco-date.pipe.spec.ts
+++ b/projects/ngneat/transloco-locale/src/lib/tests/pipes/transloco-date.pipe.spec.ts
@@ -81,9 +81,6 @@ describe('TranslocoDatePipe', () => {
 
   it('should transform an ISO 8601 string to date', () => {
     const pipe = new TranslocoDatePipe(service, cdr, LOCALE_CONFIG_MOCK);
-
-    const dateFormat: DateFormatOptions = { dateStyle: 'medium', timeStyle: 'medium' };
-    const expectedDate = Intl.DateTimeFormat('en-US', dateFormat).format(Date.UTC(2019, 8, 12, 19, 51, 33));
-    expect(pipe.transform('2019-09-12T19:51:33Z')).toEqual(expectedDate);
+    expect(pipe.transform('2019-09-12T19:51:33Z', { timeZone: 'UTC' }, 'en-US')).toEqual('Sep 12, 2019, 7:51:33 PM');
   });
 });
--- a/projects/ngneat/transloco-locale/src/lib/tests/pipes/transloco-date.pipe.spec.ts
+++ b/projects/ngneat/transloco-locale/src/lib/tests/pipes/transloco-date.pipe.spec.ts
@@ -1,6 +1,7 @@
 import { TranslocoDatePipe } from '../../pipes/transloco-date.pipe';
 import { createFakeService, createFakeCDR, LOCALE_CONFIG_MOCK } from '../mocks';
 import { defaultConfig } from '../../transloco-locale.config';
+import { DateFormatOptions } from '../../transloco-locale.types';
 
 describe('TranslocoDatePipe', () => {
   let service;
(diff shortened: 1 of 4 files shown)


shaharkazaz (approved):

projects/ngneat/transloco-locale/src/lib/tests/pipes/transloco-date.pipe.spec.ts
```diff
@@ -80,6 +81,9 @@ describe('TranslocoDatePipe', () => {
 
   it('should transform an ISO 8601 string to date', () => {
     const pipe = new TranslocoDatePipe(service, cdr, LOCALE_CONFIG_MOCK);
-    expect(pipe.transform('2019-09-12T19:51:33Z')).toEqual('Sep 12, 2019, 10:51:33 PM');
```
shaharkazaz, line 83: @chrsi Can you please add the timezone property to the `LOCALE_CONFIG_MOCK` instead of adding this code?

shaharkazaz:
@chrsi Thanks for the PR! 🙂

shaharkazaz (changes requested):
please see my comment

projects/ngneat/transloco-locale/src/lib/tests/pipes/transloco-date.pipe.spec.ts
chrsi: Do you mean extracting line 85-86 to a function in the mocks.ts file?

chrsi: There isn't really a timezone property in place, because the expected and tested value are both based on UTC.

shaharkazaz: maybe I'm missing something, what was the output before your change? and what was the cause?

chrsi: So '2019-09-12T19:51:33Z' is a date formatted in ISO 8601 representing 'Sep 12, 2019, 7:51:33 PM' (because it's represented as UTC due to the trailing Z). Once I try to transform/evaluate that datetime my machine will convert it to my current timezone (which is +2 hours) making it 'Sep 12, 2019, 9:51:33 PM'. Since you're probably running the tests in a timezone one ours ahead of mine (+3 hours) the test works fine, because the date will be transformed to 'Sep 12, 2019, 10:51:33 PM'. But for everyone that isn't in your timezone, the tests will fail.

What my code does is keeping the transformed date as is (because I'm not sure whether transloco is currently capable of setting the timezone used for interpreting the datetime) but changing the expected value based on the time zone. With my code the expected value isn't fixed to '10:51:33 PM' anymore but rather calculated from the UTC datetime under test (see `Date.UTC(2019, 8, 12, 19, 51, 33)`) on the machine that is running the tests (making it timezone aware).

I hope I could explain it clear enough :)
An alternative solution would be to make the value under test timezone aware. But I'm not sure how (or if it's possible) to set the timezone information in transloco.

chrsi: Oh ... as a matter of fact, I just found out that transloco is already capable of that. The value under test is now interpreted as a UTC datetime (because the provided ISO 8601 value uses UTC). This should make the test pass in all timezones.
Thanks for hinting me to the solution. 👍

shaharkazaz (approved):

shaharkazaz:
@chrsi the first commit was made as a fix which will bump the version, can you please update the commits to be test-related only?

chrsi:
yeah sure, done

PR #395, opened 2021-01-14
feat: pipe only loads necessary dependencies

### LINKED ISSUE #394: (Multi-scope) Transloco pipe unnecessarily loads dependencies for all scopes in ProviderScope array
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[x] Support request
[ ] Other... Please describe:
</code></pre>

## Current behavior

Currently, for implementations like `.selectTranslate()` or the Transloco template directives, a scope can be explicitly provided to determine which dependencies to load. This means when using the `multi` flag for providing scopes, using these functions or directives will only load the necessary translations.

However, the Transloco pipe does not appear to support this distinction, and will map over and load all dependencies for any scopes present in the ProviderScope array, even if only some of those translations are needed for that component.

The source code in question can be found here: https://github.com/ngneat/transloco/blob/master/projects/ngneat/transloco/src/lib/transloco.pipe.ts#L60

For example, on the `/lazy-multiple-scopes` route in the sample app, if we only want to load dependencies for one scope using the pipe, the network tab will show that additional dependencies for unused scopes have been unnecessarily fetched as well. See repro in stackblitz below.

## Expected behavior

Transloco pipe should support some mechanism of deriving the key's scope, and only loading translations for that scope. 

(PR with suggested behavior to be opened shortly.)

## Minimal reproduction of the problem with instructions

Stackblitz repro: https://stackblitz.com/edit/transloco-pipe-repro?file=src/app/lazy-multiple-scopes/lazy-multiple-scopes.component.html

To see loaded translation files, open the app in a separate window and view the Network tab in the console.

## What is the motivation / use case for changing the behavior?

Loading translations for scopes that aren't immediately needed has performance impacts, especially if those translations are coming from an external CMS.

## Environment

<pre><code>
Angular version: 7.x.x - 9.x.x

Browser:
- [x] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

kaitlynekdahl:
When using multiple scopes, Transloco pipe will derive scope from the key and only load dependencies for that scope.

Closes #394

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: https://github.com/ngneat/transloco/issues/394

When using `multi` flag, Transloco flag will map over and load dependencies for all scopes in the ProviderScope array, even if those fetched translations aren't needed for the view currently rendered. This might result in performance issues due to unnecessary API calls.

For example, the following component from the sample would only need translations for the `admin-page` scope, but the network tab shows that an unneeded call was also made for `lazy-page` translations:

image

## What is the new behavior?

The Transloco pipe should derive the scope in question from the key passed in, then determine which scope to resolve by searching the ProviderScope array for a matching scope or alias name.

After these changes, we can see that the unneeded call for `lazy-page` translations is no longer made:

image

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/ngneat/transloco/src/lib/transloco.pipe.ts
+++ b/projects/ngneat/transloco/src/lib/transloco.pipe.ts
@@ -55,11 +55,20 @@ export class TranslocoPipe implements PipeTransform, OnDestroy {
             active: activeLang
           });
 
-          return Array.isArray(this.providerScope)
-            ? forkJoin(
-                (<TranslocoScope[]>this.providerScope).map(providerScope => this.resolveScope(lang, providerScope))
-              )
-            : this.resolveScope(lang, this.providerScope);
+          let resolvedScope;
+
+          if (Array.isArray(this.providerScope)) {
+            const keyPrefix = key.split('.')[0];
+            const derivedScope = this.providerScope.find(s =>
+              typeof s === 'string' ? s === keyPrefix : s.alias === keyPrefix || s.scope === keyPrefix
+            );
+
+            resolvedScope = this.resolveScope(lang, derivedScope);
+          } else {
+            resolvedScope = this.resolveScope(lang, this.providerScope);
+          }
+
+          return resolvedScope;
         }),
         listenOrNotOperator(this.listenToLangChange)
       )


shaharkazaz:
@kaitlynekdahl Thanks for the very detailed issue 🙂 
What is the point of defining multiple scopes if they are not in use? the assumption is that the scopes defined are used by the components/sub-components and therefore we should load them.

kaitlynekdahl:
Hey @shaharkazaz, a specific scenario that I've seen is the need for all scopes to be imported at the root module level so that services that have `providedIn: root` will still be able to use Transloco and inline loaders. Because of these modules being imported at the root, any Transloco pipes at the root level will end up loading all the translations.

I think your assumption is valid in the case that an app is only using lazily-loaded modules without singleton services, though. Are you thinking the above scenario should be handled differently?

shaharkazaz:
@kaitlynekdahl Hey! sorry for the late response.
Looking again at your example, what's the benefit of setting both scopes in the providers list if not both are needed?
The loading logic is "if it's provided, it's needed", maybe the solution here should be where you are declaring your scopes.

If I misunderstood something, please feel free to share a more detailed explanation regarding your app's needs!
A dedicated repo is always the best way 😄

negue:
I'd like this feature as well. 

Currently I have components which may use different scopes inside (in order to not duplicate these strings), But adding all these scopes for each component ( / module) just takes too much time (the more components / module you split up the more scopes you need to add on each module)

Thats why I also just list all scopes in the TranslocoRootModule - and then use the strings everywhere I want.

But that way it'll load all scopes on the first resolve which you might not need (yet) 

This might be outside of the norm of Transloco but once you need to split up components into more modules (in order to create smaller bundles) the current way just adds more / repetitive work of adding/checking all scopes that are needed per component.

shaharkazaz:
@kaitlynekdahl, this will cause a side effect. For example, if you have multiple components with different scopes that should be rendered simultaneously, there could be a delay between them depending on the latency.

If you are not worried about this side effect, we'll accept a modified PR where your code is optional.

negue:
Any update on this? :)

shaharkazaz:
@negue waiting for @kaitlynekdahl to address my comments 🙂

kaitlynekdahl:
> @kaitlynekdahl, this will cause a side effect. For example, if you have multiple components with different scopes that should be rendered simultaneously, there could be a delay between them depending on the latency.
> 
> If you are not worried about this side effect, we'll accept a modified PR where your code is optional.

@shaharkazaz sorry for the even slower response! When you say there could be a delay between the components, are you saying there'd be a flash of untranslated content while the scopes are resolved synchronously?

Also for making the code optional, are you thinking there'd be a flag passed into the `params` map?

Example usage: 

```
// in the template 
<span class="admin-page-title">{{ 'AdminPageAlias.title' | transloco: {singleScope: true} }}</span>
```

Then we can use the `singleScope` param to determine whether or not to use the existing logic or the proposed logic in this PR. Does that sound like an appropriate solution?

PR #473, opened 2021-07-05
fix: allow `TranslocoService` to be garbage collected

arturovt:
Hey! I've noticed that `TranslocoService` is not getting GC'd when the app is destroyed, since cached values are capturing `this` (see code comments):

cache-leak


arturovt:
@NetanelBasal hey, would you mind having a look?

NetanelBasal:
Yes, it's no a common case, but we should handle this. @shaharkazaz, please include this in the next release.

Issue #488, opened 2021-08-16
Bug: Angular 12 full strict mode optional key pipe error


eggp:
### Current behavior
In Angular 12 full strict mode, I get an error with a conditional key.
Angular 12 tsconfig: https://github.com/eggp/translocoNG12StrictPipeError/blob/master/apps/test/tsconfig.json#L16
Error: 
```
Error: apps/test/src/app/app.component.html:1:8 - error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type 'string | null'.
  Type 'undefined' is not assignable to type 'string | null'.

1 <div>{{options1?.some | transloco}}</div>
         ~~~~~~~~~~~~~~

  apps/test/src/app/app.component.ts:6:16
    6   templateUrl: './app.component.html',
                     ~~~~~~~~~~~~~~~~~~~~~~
    Error occurs in the template of component AppComponent.
```

### Expected behavior
Accept undefined value as pipe key value.
Please add undefined to key type def. => https://github.com/ngneat/transloco/blob/faa58a5a0b73595173e3b5a640e4157d8ec50428/projects/ngneat/transloco/src/lib/transloco.pipe.ts#L35
Correct:
```
transform(key: string | null | undefined, params?: HashMap | undefined, inlineLang?: string | undefined): string {
```

### Minimal reproduction of the problem with instructions
https://github.com/eggp/translocoNG12StrictPipeError

### Environment
<pre><code>- Transloco version: 2.22.0
- Angular version: 12.X.X
</code></pre>

### Browser:
- [x] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX

Issue #311, opened 2020-07-15
transloco pipe fails with async if strict mode is enabled


artaommahe:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request
[ ] Other... Please describe:
</code></pre>

## Current behavior

Using transloco pipe after async pipe throws an exception if ts' strict mode is enabled cause transloco pipe expecting only string as input and async returns `T | null`

https://github.com/ngneat/transloco/blob/153682a4c4f3268ee20146c5f94f1baaa965008b/projects/ngneat/transloco/src/lib/transloco.pipe.ts#L34

```
Argument of type 'string | null' is not assignable to parameter of type 'string'.
Type 'null' is not assignable to type 'string'.
```

Set `"strict": false` in `tsconfig.json` and it works well.

## Expected behavior

transloco pipe works well after async pipe when strict mode is enabled

## Minimal reproduction of the problem with instructions

https://stackblitz.com/edit/transloco-async-strict

## Environment

<pre><code>
Angular version: 10.0.1
Transloco; 2.17.5

Browser:
- [x] Chrome (desktop) version 83.0.4103.116
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX

Others:

</code></pre>

shaharkazaz:
@artaommahe Hi! would you like to create a PR for it?

artaommahe:
@shaharkazaz just faced that `undefined` key value cannot be passed to `transloco` pipe with `strictTemplates: true`.

PR #497, opened 2021-09-07
Move repo to NX as part of v3 infra

shaharkazaz:
Move the repo to an NX workspace, update tests, CI, builds.


tsconfig.base.json
```diff
@@ -0,0 +1,44 @@
+{
+  "compileOnSave": false,
+  "compilerOptions": {
```
rafaelss95, line 3: Sorry to chim in, but I just noticed that the package isn't `strict` and was thinking that it could be a great opportunity to enable `"strict": true` in order to prevent mistakes like #311 and #488 :)

shaharkazaz: @rafaelss95 This is already a **very** big change, and I don't want to postpone it any further.
I'm planning to enable this in the future. This is just the first phase in several good changes 🙂

PR #579, opened 2022-07-19
fix: wrong order of language update calls

jensweigele:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The language for the parser/transpiler is set after the language of the service is updated.
The update of the language in the service triggers a rerendering - where the parser/transpiler will use the old language.
The problem is only visible if the transpiler is language-sensitive (like the MessageFormatTranspiler) since in the default case the transpiler is getting the language strings from the loader/service which already have the new language set.

Issue Number: N/A

## What is the new behavior?

The language of the transpiler is now changed before the service language is changed. This gives the language-sensitive transpilers the chance to update their data before new renderings occur. 

```
[?] Yes
[?] No
```

Not sure if this is a breaking change. It's also hard to write a test case for this usecase since it's a kind of a race-condition.

### DIFF
--- a/libs/transloco/src/lib/transloco.service.ts
+++ b/libs/transloco/src/lib/transloco.service.ts
@@ -167,8 +167,8 @@ export class TranslocoService implements OnDestroy {
   }
 
   setActiveLang(lang: string) {
-    this.lang.next(lang);
     this.parser.onLangChanged?.(lang);
+    this.lang.next(lang);
     this.events.next({
       type: 'langChanged',
       payload: getEventPayload(lang),


ivnbogdan:
Found this same issue in our project.
Logged an issue about it here: https://github.com/ngneat/transloco/issues/610

Would love to get this merged ❤️  
Was going to make the same PR and found this instead. Issue is that after language change, first parsing is done with the old parser option. This current fix updates the options first, so parsing is done correctly from the start.

I'm not a contributor, but @jensweigele might want to rename the commit message with something like `fix(messageformat): update order of operations on language change`. Just a thought 🤞

ben12:
Same code fix as #634, already merged.

shaharkazaz:
@ben12 is correct. Thanks @jensweigele for the PR. this code is available in v4.2.6.

PR #624, opened 2023-02-09
fix(migrate): continue with null value instead of  TypeError: Cannot read property 'trim' of undefined.

fowl2:
This can happen in the presence of void elements.

## PR Checklist

## PR Type

```
[x] Bugfix
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/schematics/src/ng-migrate/ng-migrate.ts
+++ b/schematics/src/ng-migrate/ng-migrate.ts
@@ -77,7 +77,12 @@ function getTranslation(template) {
     }
 
     key = kebabCase(key);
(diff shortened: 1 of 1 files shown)


shaharkazaz:
@fowl2 Thanks for the PR, what was the case that you encountered that?

fowl2:
IIRC, I had a void element with a i18n attribute. An <input> I think. Technically illegal html, but it was there and stopped the whole migration.

shaharkazaz:
Can you please add a comment with the case? to make this change clearer

shaharkazaz (approved):

PR #639, opened 2023-03-23
perf(transloco): use EmbeddedViewRef to detect changes

edbzn:
## PR Checklist

## PR Type

Performance-related refactoring.

```
[ ] Bugfix
[ ] Feature
[x] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The current behavior is using ChangeDetectorRef#markForCheck that relies on Zone.js and marks the component and its parents as dirty, waiting for the Zone to trigger change detection. This is especially bad for leaf components as it will mark the whole component tree as dirty before being able to update the desired template.

## What is the new behavior?

Uses EmbeddedViewRef#detectChanges only checks the view and its children, which is the smallest unit in the change detection process of a component's template, which gives a performance boost and works zoneless.  

```
[ ] Yes
[x] No
```

### DIFF
--- a/libs/transloco/src/lib/transloco.directive.ts
+++ b/libs/transloco/src/lib/transloco.directive.ts
@@ -1,5 +1,4 @@
 import {
-  ChangeDetectorRef,
   Directive,
   ElementRef,
   EmbeddedViewRef,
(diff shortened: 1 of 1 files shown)


shaharkazaz:
@edbzn Thanks for the PR!
I think that's a change that makes sense to do in a zoneless Angular. In the current state of things, it will likely cause redundant CD cycles instead of improving the performance.
Here's why:
1. If a child component is calling `markForCheck` then we will force another CD for no reason.
2. Almost every event within the nested tree will cause a check on all the parent components which is highly likely to happen, so we want to tap into this already existing check rather than force an immediate CD.

I do agree that Angular is moving towards a zoneless future and this change is likely to be part of it, but as of now, I don't find this change relevant.

shaharkazaz:
@edbzn if you wan to talk about creating something for zoneless applications (maybe a TranslocoSignalService) DM me.
Closing this for now.

PR #641, opened 2023-03-28
feat(transloco-locale): expose default transformers (#435)

leptoquark1:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[x] Documentation content changes
[ ] Other... Please describe:
```

When implementing custom transformers, you can only fallback to the default behaviour by code duplicating code from not exported internal functions of transloco-locale

Issue Number: N/A
Discussion Number: #435

## What is the new behavior?

`DefaultDateTransformer` and `DefaultNumberTransformer` are now exposed to public api, so that they can be referenced and used as fallback in custom implementations.

```
[ ] Yes
[x] No
```

### DIFF
--- a/docs/docs/plugins/locale.mdx
+++ b/docs/docs/plugins/locale.mdx
@@ -280,13 +280,18 @@ In some cases you might want to customize the localization transformation, to ac
 There are two types of transformers for date and number:
 
 ```ts title="custom-transformer.ts"
-export class CustomDateTransformer implements TranslocoDateTransformer {
+import { DefaultDateTransformer, DefaultNumberTransformer } from '@ngneat/transloco-locale';
+
+export class CustomDateTransformer extends DefaultDateTransformer {
   public transform(date: Date, locale: Locale, options: DateFormatOptions): string {
     return ...
+
+    // Fallback to default transformer
+    return super.transform(date, locale, options);
   }
 }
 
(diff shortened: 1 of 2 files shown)


leptoquark1:
@shaharkazaz Ping, in favor of #435

docs/docs/plugins/locale.mdx
shaharkazaz, line 1: I don't understand why do we need this doc change 🤔

shaharkazaz (commented):
Thanks for the PR 👍

docs/docs/plugins/locale.mdx
leptoquark1: Just a suggestion. I think most implementations of custom transformers make use of the default value as fallback

shaharkazaz: Ok, please add the import of these transformers as well.

The usage of: `DefaultDateTransformer(...args)`  is incorrect. this is a class and the current use should be something like this:
```
export class CustomDateTransformer extends DefaultDateTransformer implements TranslocoDateTransformer {
  public transform(date: Date, locale: Locale, options: DateFormatOptions): string {
    return ...
    ...
    // Fallback to default transformer
    return super.transform(date, locale, options);
  }
}
```

PR #650, opened 2023-04-26
feat(transloco): support nested array of scopes

dmitry-stepanenko:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Right now transloco is able to resolve only 1-level array of scopes. This works fine with straightforward approach when using `{ provide: TRANSLOCO_SCOPE, useValue: scope, multi: true }`. However, it might be needed to pass an array of scopes within single provider.

## What is the new behavior?
This change allows to use the following syntax
```
{ 
  provide: TRANSLOCO_SCOPE,
  useValue: ['scope1', 'scope2']
  multi: true
}
```

In my scenario there's a big amount of scopes that we're using. They're provided either as component or module providers. And there's a limitation related to Angular's injection hierarchy: it resolves ElementInjectors before ModuleInjectors and if TRANSLOCO_SCOPE is provided on the ElementInjector level, it won't grab module injectors at all. To overcome this I want to grab all parent scopes and provide them explicitly

```
export function provideLanguageScopes(
  scope: LanguageScope,
  ...otherScopes: LanguageScope[]
): Provider {
  const scopeSet = new Set(otherScopes.concat(scope));
  return {
    provide: TRANSLOCO_SCOPE,
    useFactory: (parentScopes: LanguageScope[] | null) => {
      parentScopes && parentScopes.flat(Infinity).forEach(p => scopeSet.add(p));
      return [...scopeSet.values()];
    },
    multi: true,
    deps: [[new SkipSelf(), new Optional(), TRANSLOCO_SCOPE]],
  };
}
```

This approach requires transloco to be able to work with nested arrays of scopes. This PR does not introduce any changes to existing behavior, only enhances it so that it is possible to provide arrays under `TRANSLOCO_SCOPE`

```
[x] Yes
[ ] No
```
Setting `yes` as there's a change to `MaybeArray` interface, that might not be compatible with all implementations

### DIFF
--- a/libs/transloco/src/lib/scope-resolver.ts
+++ b/libs/transloco/src/lib/scope-resolver.ts
@@ -1,10 +1,10 @@
-import { TranslocoScope, ProviderScope, MaybeArray } from './types';
+import { TranslocoScope, ProviderScope } from './types';
 import { TranslocoService } from './transloco.service';
 import { isScopeObject, toCamelCase } from './helpers';
 
 type ScopeResolverParams = {
   inline: string | undefined;
-  provider: MaybeArray<TranslocoScope>;
+  provider: TranslocoScope;
 };
 
 export class ScopeResolver {
--- a/libs/transloco/tsconfig.lib.json
+++ b/libs/transloco/tsconfig.lib.json
@@ -7,7 +7,7 @@
     "declarationMap": true,
     "inlineSources": true,
     "types": [],
-    "lib": ["dom", "es2018"]
+    "lib": ["dom", "es2019"]
   },
   "exclude": ["src/test-setup.ts", "**/*.spec.ts"],
   "include": ["**/*.ts"]
--- a/libs/transloco/src/lib/helpers.ts
+++ b/libs/transloco/src/lib/helpers.ts
@@ -1,4 +1,4 @@
-import { ProviderScope, Translation } from './types';
+import { NestedArray, ProviderScope, Translation } from './types';
 import { flatten as _flatten, unflatten as _unflatten } from 'flat';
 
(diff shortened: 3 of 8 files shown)


libs/transloco/src/lib/scope-resolver.ts
```diff
@@ -1,10 +1,10 @@
-import { TranslocoScope, ProviderScope, MaybeArray } from './types';
+import { TranslocoScope, ProviderScope } from './types';
 import { TranslocoService } from './transloco.service';
 import { isScopeObject, toCamelCase } from './helpers';
 
 type ScopeResolverParams = {
   inline: string | undefined;
-  provider: MaybeArray<TranslocoScope>;
+  provider: TranslocoScope;
```
dmitry-stepanenko, line 7: I believe this was done unintentionally, scope resolver is not able to process array of scopes. There's no compilation error only because of `isScopeObject` type guard, that is also not supposed to handle array of scopes

libs/transloco/tsconfig.lib.json
```diff
@@ -7,7 +7,7 @@
     "declarationMap": true,
     "inlineSources": true,
     "types": [],
-    "lib": ["dom", "es2018"]
+    "lib": ["dom", "es2019"]
```
dmitry-stepanenko, line 10: to be able to use `Array.prototype.flat()`

dmitry-stepanenko:
@shaharkazaz will you be able to look at this? An issue on our side appears to be quite serious

PR #689, opened 2023-08-10
feat(transloco): 🎸 allow to omit available languages

### LINKED ISSUE #653: Feature(config): support translation without needing to specify availableLangs

- [X] I have searched the existing issues

### Which Transloco package(s) will this feature affect?

Transloco

### Is your feature request related to a problem? Please describe

In our company, we have a third party tool where the translations are maintained. The developers add the keys in the english language and upload the keys to the tool. Someone else will take care that the keys will be translated in all supported languages. During build, we download all the translation files. However, the developers don't care which languages are supported, either the translation file is there, or it's missing and we run the fallback logic. Transloco though forces us developers to maintain a list of all supported languages, which is annoying for us as it's not our responsibility. Adding new languages now requires code changes, which complicates this process unnecessarily.

### Describe the solution you'd like

In `TranslocoConfig`, when omitting `availableLangs`, or setting it to `null`, I can load any language without restrictions.

### Describe alternatives you've considered

We can bite the bullet and maintain the list of supported languages

### Additional context

We are migrating from ngx-translate, there we don't need to specify the available languages.

### I would like to make a pull request for this feature

Yes 🚀

json-derulo:
Currently the available languages need to be set, otherwise the translated values won't be returned. With this change the translation will also work when omitting the availableLangs config, or setting it to null.

✅ Closes: #653

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently it is not allow to omit `availableLangs` by typings. When setting it to empty array, translation will never work, because there is a check whether the current language is available. There is no way to skip this check.

Issue Number: #653

## What is the new behavior?

The check whether the current language is in the list of available languages can be skipped by omitting `availableLangs` in the config or setting it to `null`. This allows translation without the need of maintaining `availableLangs`.

- [ ] Yes
- [x] No

### DIFF
--- a/docs/docs/getting-started/config-options.mdx
+++ b/docs/docs/getting-started/config-options.mdx
@@ -60,6 +60,8 @@ translocoConfig({
   availableLangs: ['en', 'es']
 })
 ```
+When the active language is not included in the `availableLangs`, the translations of this language will not be returned.
+You can omit this option or set it to `null`, in this case the translations will be returned for any language, if available.
 
 ### `missingHandler.allowEmpty`
 Whether to allow empty values: (defaults to `false`)
--- a/libs/transloco/src/lib/tests/service/translate.spec.ts
+++ b/libs/transloco/src/lib/tests/service/translate.spec.ts
@@ -79,4 +79,16 @@ describe('translate', () => {
       'Admin Lazy spanish'
     );
   }));
+  
+  it('should return the translation when availableLangs is skipped', fakeAsync(() => {
+    service.setAvailableLangs(undefined);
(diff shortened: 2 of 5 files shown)


shaharkazaz:
Thanks for the PR!
I have to say that I don't have full confidence in this solution as it feels more like shortcircuiting the logic that was made there.
We need to understand how this affects the flow of things as this was made for a reason

lppedd:
I actually like this. The less configuration, the better imo.

Issue #708, opened 2023-10-01
Feature(transloco): Allow to provide multiple scope at once


EricPoul:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Which Transloco package(s) will this feature affect?

Transloco

### Is your feature request related to a problem? Please describe

I have server-side translations that are divided by scope/interface but also there are shared scopes that contain translations reused through the app, so I have to provide several scopes for lazy-loaded features.
Currently, it's possible to provide multiple scopes but the type says that only one is possible to be provided. Also, `provideTranslocoScope` allows to provide only one scope so I have to provide `provideTranslocoScope` multiple time that is frustrating and bloats the code.

### Describe the solution you'd like

It'd be great if we could do the next:
```ts
provideTranslocoScope('core', 'todos');
```
instead of:
```ts
provideTranslocoScope('core');
provideTranslocoScope('todos');
```

### Describe alternatives you've considered

```markdown
-
```

### I would like to make a pull request for this feature

Yes 🚀

PR #715, opened 2023-10-14
feat(transloco): allow to provide multiple scopes via `provideTranslocoScopes`

EricPoul:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/ngneat/transloco/issues/708

- [x] Yes
- [ ] No

### DIFF
--- a/libs/transloco/CHANGELOG.md
+++ b/libs/transloco/CHANGELOG.md
@@ -2,25 +2,26 @@
 
 This file was generated using [@jscutlery/semver](https://github.com/jscutlery/semver).
 
-# [7.0.0](https://github.com/jsverse/transloco/compare/transloco-6.0.4...transloco-7.0.0) (2024-03-23)
-
-
-### chore
-
-* **transloco:** 🤖 update package scope ([f8b6d58](https://github.com/jsverse/transloco/commit/f8b6d58f51d99fcb2912082fbefd95527a57e2e4))
-
+# [7.1.0](https://github.com/ngneat/transloco/compare/transloco-7.0.0...transloco-7.1.0) (2024-03-30)
 
 ### Features
 
-* **transloco:** 🎸 Improve params algo matching ([8447634](https://github.com/jsverse/transloco/commit/8447634588210771047c45072186f20867fd8a2f))
+- **transloco:** 🎸 introduce prefix input ([1e27a8f](https://github.com/ngneat/transloco/commit/1e27a8fe9922e3e8f4ced0ea3960aca8ae4da92d))
 
+# [7.0.0](https://github.com/jsverse/transloco/compare/transloco-6.0.4...transloco-7.0.0) (2024-03-23)
+
+### chore
+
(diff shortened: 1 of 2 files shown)


EricPoul:
@NetanelBasal could someone check this one?

alexfriesen:
This looks like a breaking change as you are removing `provideTranslocoScope`. I think adding `provideTranslocoScopes` as an additional function would be a better idea

EricPoul:
I don't mind adding a new function and the previous one marking as deprecated

farshadhemmati (approved):

EricPoul:
@farshadhemmati would you like to keep an old `provideTranslocoScope` in addition to `provideTranslocoScopes` or just a new `provideTranslocoScopes` with breaking changes?

libs/transloco/src/lib/transloco.providers.ts
```diff
@@ -76,12 +76,12 @@ export function provideTranslocoLoader(loader: Type<TranslocoLoader>) {
   ]);
 }
 
-export function provideTranslocoScope(scope: TranslocoScope) {
-  return {
+export function provideTranslocoScopes(...scopes: TranslocoScope[]) {
```
shaharkazaz, line 79: keep the original name `provideTranslocoScope` and let's merge this, no need to make this a breaking change over a name as all the previous usages will still work as expected (correct my if I'm wrong)

shaharkazaz:
@EricPoul We can merge this let's just skip the function rename

majkers:
@EricPoul @shaharkazaz Is it just me or there is something wrong with it now?

I have:

```
@Component({
...
  providers: [provideTranslocoScope('ea')],
  standalone: true,
...
})
```

and then:

`private scope: TranslocoScope = inject(TRANSLOCO_SCOPE);
`
and when calling 

`this.translocoService.selectTranslate('header.logOut', undefined, this.scope)
`
I get an error because it is an array now:

```
core.mjs:6531 ERROR TypeError: lang.split is not a function
    at getLangFromScope (jsverse-transloco.mjs:379:17)
    at TranslocoService.isScopeWithLang (jsverse-transloco.mjs:658:28)
    at TranslocoService.selectTranslate (jsverse-transloco.mjs:642:39)
```

but when I go back to decalre scope like this:

it is working OK

```
providers: [{
    provide: TRANSLOCO_SCOPE,
    useValue: 'ea'
  }],
```

shaharkazaz:
@majkers hm... It's a case that got missed when adding this feature, since now the scope is provided as multi the inject will return an array which I guess is a breaking change.

As a workaround you can just access the scope at position 0.

A proposed fix might be to drop the multi flag in case one scope is provided, but the thing with this is that you can get either an array or a scope if you inject it.

WDYT? @EricPoul

shaharkazaz:
@majkers thanks for the extra info!
Would you like to open a PR for the fix? Or I'll take it?

shaharkazaz:
@majkers There is always a first time! I think it's a nice chance to contribute to an open source, especially after you have already collected all the information needed 😀 

But of course it's your call 🙏

shaharkazaz:
@majkers should be ok in v7.4.1

PR #754, opened 2024-03-22
Add the possibility to bypass the cached values to load new translation

simon-knu:
This pull request addresses the open issue https://github.com/jsverse/transloco/issues/474.


JonasDev17:
Would be nice if this were to be merged cause after every update we publish, if the user still has an open session he will see untranslated strings if he does not reload the page which is not ideal. It'd be nice to be able to force reload the translation.

PR #760, opened 2024-04-23
fix(transloco): fixed error when inline loader for lang is not provided

adpawlik:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

If you use an inline loader and specify a language that is not included in the map, an error occurs.
e.g.
```
export const loader = ['en', 'es'].reduce((acc, lang) => {
    acc[lang] = () => import(`../i18n/${lang}.json`);
    return acc;
}, {});

export const FEATURE_ROUTES: Route = {
  path: 'feature',
  loadComponent: () => import('./feature.component').then(FeatureComponent => FeatureComponent),
  providers: [
    provideTranslocoScope({
      scope: 'scopeName',
      loader
    })
  ]
};
```
Issue Number: N/A

## What is the new behavior?
The lack of a language in the given map does not cause the application to crash, it only logs an error in the case of the development version and loads a fallback language

- [ ] Yes
- [x] No

### DIFF
--- a/libs/transloco/src/lib/resolve-loader.ts
+++ b/libs/transloco/src/lib/resolve-loader.ts
@@ -15,7 +15,7 @@ export function resolveLoader(options: Options) {
   if (inlineLoader) {
     const pathLoader = inlineLoader[path];
     if (isFunction(pathLoader) === false) {
-      throw `You're using an inline loader but didn't provide a loader for ${path}`;
+      return Promise.reject(`You're using an inline loader but didn't provide a loader for ${path}`);
     }
 
     return inlineLoader[path]().then((res) =>


wszydlak (approved):

shaharkazaz:
@adpawlik I'm not sure that crashing the app is the desired behavior here, why did you feel you needed to make this change?

adpawlik:
@shaharkazaz Because it is not good to throw an error at this point, because the "resolveLoader" function is used in transloco.service.ts and such an error will not be caught by catchError in rxjs, and therefore the fallback language will not be loaded.

Currently, all inline loaders must have map entries for each available language.

e.g.
```
//componentA 
export const loader = ['en', 'es'].reduce((acc, lang) => {
    acc[lang] = () => import(`../i18n/${lang}.json`);
    return acc;
}, {});

//componentB 
export const loader = ['en', 'es', 'de'].reduce((acc, lang) => {
    acc[lang] = () => import(`../i18n/${lang}.json`);
    return acc;
}, {});
```
If you try to load the 'de' language for componentA, it fails.

Issue #755, opened 2024-04-03
Feature(transloco|transloco-locale): Provide complete type definitions for pipes used in strict mode


Kaemmelot:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Which Transloco package(s) will this feature affect?

Transloco, Locale

### Is your feature request related to a problem? Please describe

Only the transloco pipe allows to pass `undefined` or `null` as input values in strict mode. Other pipes like the date pipe or decimal pipe cannot be used with these values. Additionally, the transloco pipe does not provide the types returned by the `TranslocoMissingHandler`, so returning `undefined` there is also a problem in strict mode.

### Describe the solution you'd like

I think each pipe should have multiple overloads that consider `null` and `undefined`:
```ts
transform(value: TYPE): string;
transform(value: TYPE | null | undefined): string | null | undefined;
transform(value: null | undefined): null | undefined;
```
The transloco pipe should probably provide similar overloads. But I'm not sure about the return value for that pipe, because the `TranslocoMissingHandler` can return `any`, which probably needs to be restricted before providing better return types for that pipe. That would be a breaking change.

I would also return the unchanged input value if it was `null` or `undefined`. This would not print anything if used in a template, but would allow to use the correct typescript features to check for these input values based on the result and without introducing a variable (for example if the input is the output of another pipe):
```ts
// pipe returning empty string:
{{ (value | somePipe | translocoDecimal) || 'N/A' }}
// pipe returning null or undefined:
{{ (value | somePipe | translocoDecimal) ?? 'N/A' }}
```

### Describe alternatives you've considered

```markdown
I have looked at how the Angular team provided their pipes, like for example the [date pipe](https://github.com/angular/angular/blob/6f960afd31943a1316c0dab7a7659f9bd9a14739/packages/common/src/pipes/date_pipe.ts#L251) or the [decimal pipe](https://github.com/angular/angular/blob/6f960afd31943a1316c0dab7a7659f9bd9a14739/packages/common/src/pipes/number_pipe.ts#L86) which is one reason for this ticket.

I also saw that they only provide `null` as return value, because they could not add `undefined` later on as described [here](https://github.com/angular/angular/pull/39785#issuecomment-731418991), so I don't think transloco should follow that example for the return values.
```

### Additional context

Related to #311 and #488

### I would like to make a pull request for this feature

Yes 🚀

shaharkazaz:
You are welcome to open a PR for it 👍

PR #770, opened 2024-06-06
feat(transloco|locale): Improve pipe type definitions for strict mode

Kaemmelot:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently the pipes don't provide overloads that can return null or undefined if provided in the input. The locale pipes transform these values into empty strings, which makes `value | pipe ?? elseValue` use cases hard. The typings are also not helpful for strict mode.

Issue Number: #755 

## What is the new behavior?

The pipes now return null or undefined as given in the input and provide method overloads to keep the typings intact for strict mode.

- [x] Yes
- [ ] No

These changes could potentially be breaking in some cases as the locale pipes now return null or undefined where they previously returned an empty string instead. The transloco pipe has new return types but returns the same values as previously, so I don't think that pipe has a breaking change.

## Other information

The order of the overloads is important, because typescript tries to use the first matching overload. The docs can be seen for every overload if provided on the first one, so they must be on top.

### DIFF
--- a/libs/transloco/CHANGELOG.md
+++ b/libs/transloco/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 7.6.1 (2025-03-24)
+
+
+### 🩹 Fixes
+
+- **transloco:** 🐛 ngDevMode declaration issue ([#843](https://github.com/jsverse/transloco/pull/843))
+
 ## 7.6.0 (2025-03-22)
 
 
--- a/libs/transloco/package.json
+++ b/libs/transloco/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@jsverse/transloco",
-  "version": "7.6.0",
+  "version": "7.6.1",
   "description": "The internationalization (i18n) library for Angular",
   "schematics": "./schematics/src/collection.json",
   "publishConfig": {


libs/transloco-locale/src/lib/pipes/transloco-decimal.pipe.ts
```diff
@@ -31,12 +31,39 @@ export class TranslocoDecimalPipe
...
+    locale?: Locale
+  ): string | undefined;
+  transform(
+    value: number | string | null | undefined,
+    numberFormatOptions?: NumberFormatOptions,
+    locale?: Locale
+  ): string | null | undefined;
+
+  transform(
+    value?: string | number | null,
```
shaharkazaz, line 62: The overloading seems robust.
Why not simply:
```ts
  transform(
      value: string | number,
      numberFormatOptions?: NumberFormatOptions,
      locale?: Locale
  ): string;
  transform<T extends undefined | null>(
      value: T,
      numberFormatOptions?: NumberFormatOptions,
      locale?: Locale
  ): T;
  transform(
    value: string | number | undefined | null,
    numberFormatOptions: NumberFormatOptions = {},
    locale?: Locale
  )
```
Seems like to cover the cases corretly

shaharkazaz:
@Kaemmelot Did you happen to see my comment?

libs/transloco-locale/src/lib/pipes/transloco-decimal.pipe.ts
Kaemmelot: Don't know why I didn't think of this. I adjusted the commits, thanks.

Kaemmelot:
@shaharkazaz I saw your comment and was about to change the commits.

Since this is the first time I wanted to rebase changes with GitHub as a tool, I tried the Sync button in my fork. That added an unwanted merge commit. Guess I'm not clicking that again and continue with Git and force push. Haven't used GitHub that much, some parts of the UI are a bit confusing I think.

libs/transloco-locale/src/lib/pipes/transloco-currency.pipe.ts
```diff
@@ -35,14 +37,37 @@ export class TranslocoCurrencyPipe
...
+  ): string;
+  transform<T extends null | undefined>(
+    value: T,
+    display?: CurrencyDisplayType,
+    numberFormatOptions?: NumberFormatOptions,
+    currencyCode?: Currency,
+    locale?: Locale
+  ): T;
+  transform<T extends null | undefined>(
+    value: number | string | T,
```
shaharkazaz, line 56: My suggestion didn't work?

Kaemmelot: Not sure if I understood you correctly. I guess you meant to remove the last overload? If I would remove this overload from this pipe and pass in a `string | null` value, I would get this error:

> No overload matches this call.
  Overload 1 of 2, '(value: string | number, display?: CurrencyDisplayType | undefined, numberFormatOptions?: NumberFormatOptions | undefined, currencyCode?: string | undefined, locale?: string | undefined): string', gave the following error.
    Argument of type 'string | null' is not assignable to parameter of type 'string | number'.
      Type 'null' is not assignable to type 'string | number'.
  Overload 2 of 2, '(value: null | undefined, display?: CurrencyDisplayType | undefined, numberFormatOptions?: NumberFormatOptions | undefined, currencyCode?: string | undefined, locale?: string | undefined): null | undefined', gave the following error.
    Argument of type 'string | null' is not assignable to parameter of type 'null | undefined'. ts(2769)
transloco-currency.pipe.ts(56, 3): The call would have succeeded against this implementation, but implementation signatures of overloads are not externally visible.

So I think I need all three overloads.

PR #794, opened 2024-09-12
feat(transloco): 🎸 add option to not change the casing of the provided scope

maartentibau:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Documentation content changes

Currently the `scope` is being altered by the `toCamelCase` helper method. This creates the issue that if you provide a `scope` where you would like to have the first character not to be changed OR just the casing in general you can only do this by providing an `alias`.

e.g `{ scope: 'FOOBAR', alias: 'FOOBAR' }`

Issue Number: N/A

## What is the new behavior?

This feature enabled the possibility to only provide the scope and the casing of it will not be altered, so you can just use this: `{ scope: 'FOOBAR' }` or `{ scope: 'F-OO-BaR' } --> will be transformed to `{ scope: 'FOOBaR' }` 

- [ ] Yes
- [x] No

### DIFF
--- a/docs/docs/getting-started/config-options.mdx
+++ b/docs/docs/getting-started/config-options.mdx
@@ -133,3 +133,14 @@ translocoConfig({
   interpolation: ['<<<', '>>>'],
 });
 ```
+
+### `scopes.keepCasing`
+This will make sure that the casing of the provided scope name is not altered, given that no alias has been set otherwise this setting will be ignored.
+
+```ts
+translocoConfig({
+  scopes: {
+    keepCasing: false
+  }
+})
+```
--- a/docs/docs/lazy-load/scope-configuration.mdx
+++ b/docs/docs/lazy-load/scope-configuration.mdx
@@ -139,8 +139,9 @@ Now we can access each one of the `todos` keys by using the `todos` namespace:
 ```
 
 ## Scope's namespace
+By default, the namespace will be the `scope`'s name camel-cased. you can keep the original casing by providing the [`scope.keepCasing`](../getting-started/config-options#scopeskeepcasing) config option.
 
-By default, the namespace will be the `scope` name (camel cased), but we can override it by providing custom `alias` name in the module/component `scope` provider:
+You can also provide a custom scope namespace by specifying an `alias` name in the module/component `scope` provider:
 
 ```ts
 provideTranslocoScope({ scope: 'todos', alias: 'customName' });
--- a/libs/transloco/src/lib/scope-resolver.ts
+++ b/libs/transloco/src/lib/scope-resolver.ts
@@ -19,7 +19,12 @@ export class ScopeResolver {
 
     if (provider) {
       if (isScopeObject(provider)) {
-        const { scope, alias = toCamelCase(scope) } = provider as ProviderScope;
+        const {
+          scope,
+          alias = this.service.config.scopes.keepCasing
+            ? scope
+            : toCamelCase(scope),
+        } = provider as ProviderScope;
         this.service._setScopeAlias(scope, alias);
 
         return scope;
--- a/libs/transloco/src/lib/tests/scope-resolver.spec.ts
+++ b/libs/transloco/src/lib/tests/scope-resolver.spec.ts
@@ -8,6 +8,11 @@ describe('ScopeResolver', () => {
     spy = jasmine.createSpy('setScopeAlias');
     resolver = new ScopeResolver({
       _setScopeAlias: spy,
+      config: {
+        scopes: {
+          keepCasing: false,
+        },
+      },
     } as any);
   });
 
@@ -81,5 +86,21 @@ describe('ScopeResolver', () => {
       'nested/scopes/admin-page',
       'nestedScopesAdminPage',
     );
+  });
+
+  it('should keep the original scope name', () => {
+    resolver = new ScopeResolver({
+      _setScopeAlias: spy,
+      config: { scopes: { keepCasing: true } },
+    } as any);
+
+    expect(
+      resolver.resolve({
+        inline: undefined,
+        provider: { scope: 'AdMiN-pAgE' },
+      }),
+    ).toEqual('AdMiN-pAgE');
+    expect(spy).toHaveBeenCalledTimes(1);
+    expect(spy).toHaveBeenCalledWith('AdMiN-pAgE', 'AdMiN-pAgE');
   });
 });
--- a/libs/transloco/src/lib/transloco.service.ts
+++ b/libs/transloco/src/lib/transloco.service.ts
@@ -789,8 +789,10 @@ export class TranslocoService implements OnDestroy {
   }
 
   private getMappedScope(scope: string): string {
-    const { scopeMapping = {} } = this.config;
-    return scopeMapping[scope] || toCamelCase(scope);
+    const { scopeMapping = {}, scopes = { keepCasing: false } } = this.config;
+    return (
+      scopeMapping[scope] || (scopes.keepCasing ? scope : toCamelCase(scope))
+    );
   }
 
   /**
--- a/libs/transloco/src/lib/tests/service/setTranslation.spec.ts
+++ b/libs/transloco/src/lib/tests/service/setTranslation.spec.ts
@@ -102,5 +102,16 @@ describe('setTranslation', () => {
       };
       expect(setTranslationsSpy).toHaveBeenCalledWith('en', merged);
     });
+
+    it("should not change scope's name given scope.keepCasing is set to true", () => {
+      service.config.scopes.keepCasing = true;
+      lang = 'LAZY-page/en';
+      service.setTranslation(translation, lang);
+      const merged = {
+        ...flatten(mockLangs.en),
+        ...flatten({ 'LAZY-page': { ...translation } }),
+      };
+      expect(setTranslationsSpy).toHaveBeenCalledWith('en', merged);
+    });
   });
 });
--- a/libs/transloco/src/lib/transloco.config.ts
+++ b/libs/transloco/src/lib/transloco.config.ts
@@ -18,6 +18,9 @@ export interface TranslocoConfig {
     allowEmpty: boolean;
   };
   interpolation: [string, string];
+  scopes: {
+    keepCasing?: boolean;
+  };
 }
 
 export const TRANSLOCO_CONFIG = new InjectionToken<TranslocoConfig>(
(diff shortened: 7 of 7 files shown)


libs/transloco/src/lib/scope-resolver.ts
```diff
@@ -8,7 +15,14 @@ type ScopeResolverParams = {
...
 
 export class ScopeResolver {
-  constructor(private service: TranslocoService) {}
+  readonly config: TranslocoConfig;
+
+  constructor(
+    private service: TranslocoService,
+    @Inject(TRANSLOCO_CONFIG) userConfig: PartialTranslocoConfig = {},
+  ) {
+    this.config = JSON.parse(JSON.stringify(userConfig));
```
shaharkazaz, line 24: No need for the clone, we are just reading properties.

libs/transloco/src/lib/scope-resolver.ts
```diff
@@ -8,7 +15,14 @@ type ScopeResolverParams = {
 };
 
 export class ScopeResolver {
-  constructor(private service: TranslocoService) {}
+  readonly config: TranslocoConfig;
+
+  constructor(
+    private service: TranslocoService,
+    @Inject(TRANSLOCO_CONFIG) userConfig: PartialTranslocoConfig = {},
```
shaharkazaz, line 22: You can't use angular injection as this is a pure JS class and isn't an injectable.

shaharkazaz (commented):
Thanks for the PR, nice feature! left a couple of comments 🙂

libs/transloco/src/lib/helpers.ts
```diff
@@ -86,10 +86,14 @@ export function coerceArray<T>(value: T | T[]): T[] {
  * given: path-to_happiness => pathToHappiness
  *
  */
-export function toCamelCase(str: string): string {
+export function toCamelCase(str: string, ignoreLowerUpperCase = false): string {
```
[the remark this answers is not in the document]
maartentibau: Here I was a bit in doubt myself to be honest.

Because this method does not only do a camel casing but also replaces, dashes, underscorse, forward slashes, spaces... so i was a bit confused.

So I added this to comply with the fact we only ignore the casing, but I think it would indeed be better to just let it up to the consumer.

But then the next question might rise, is `keepNamespaceCasing` still a good name?
Maybe we could go for `scope.keepNamespace` of `scope.noCamelCase` ?

libs/transloco/src/lib/scope-resolver.ts
maartentibau: Fixed

libs/transloco/src/lib/scope-resolver.ts
maartentibau: Removed, used the `config` from the service

maartentibau:
I will squash all the commits once we get to a point where the PR Is final, but I now left them there so you can see changes a bit more easy.

docs/docs/getting-started/config-options.mdx
```diff
@@ -113,3 +113,12 @@ translocoConfig({
   interpolation: ['<<<', '>>>']
 })
 ```
+
+### `ignoreScopeCase`
+The start and end markings for parameters: (defaults to `['{{', '}}']`)
```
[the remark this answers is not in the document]
maartentibau: Done

docs/docs/getting-started/config-options.mdx
```diff
@@ -113,3 +113,12 @@ translocoConfig({
...
+```ts
+translocoConfig({
+  // This will make sure that the casing of the scope provided is not altered, given that no alias has been set otherwise this setting will be ignored
+  ignoreScopeCase: true
```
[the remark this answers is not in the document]
maartentibau: I was not 100% sure what you meant by this, I assumed

```json
scope: {
  keepNamespaceCasing
}
```

libs/transloco/src/lib/helpers.ts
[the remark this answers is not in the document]
shaharkazaz: I find `keepNamespaceCasing` very explicit. Maybe `scope.keepCasing`.

docs/docs/getting-started/config-options.mdx
```diff
@@ -113,3 +113,13 @@ translocoConfig({
   interpolation: ['<<<', '>>>']
 })
 ```
+
+### `scopes.keepNamespaceCasing`
+```ts
+translocoConfig({
+  // This will make sure that the casing of the provided scope name is not altered, given that no alias has been set otherwise this setting will be ignored
```
shaharkazaz, line 120: Move this comment to be text under the title.

docs/docs/lazy-load/scope-configuration.mdx
```diff
@@ -140,7 +140,9 @@ Now we can access each one of the `todos` keys by using the `todos` namespace:
 
 ## Scope's namespace
 
-By default, the namespace will be the `scope` name (camel cased), but we can override it by providing custom `alias` name in the module/component `scope` provider:
+By default, the namespace will be the `scope` name (camel cased), but we can override it by setting the `scope.keepNamespaceCasing` option to `true` in the `config`.
```
shaharkazaz, line 143:
```suggestion
By default, the namespace will be the `scope`'s name camel-cased. you can keep the original casing by providing the [`scope.keepNamespaceCasing`](../getting-started/config-options#scopeskeepnamespacecasing) config option.

You can also provide a custom scope namespace by specifying an `alias` name in the module/component `scope` provider:
```

Just verify that the link works

libs/transloco/src/lib/tests/scope-resolver.spec.ts
```diff
@@ -73,10 +73,36 @@ describe('ScopeResolver', () => {
...
+    } as any);
+
+    expect(
+      resolver.resolve({
+        inline: undefined,
+        provider: {
+          scope: 'AdMiN-pAgE',
+        },
+      }),
+    ).toEqual('AdMiN-pAgE');
```
shaharkazaz, line 103: WDYT, should we protect from invalid chars?

libs/transloco/src/lib/tests/scope-resolver.spec.ts
```diff
@@ -73,10 +73,36 @@ describe('ScopeResolver', () => {
...
     // one from before
     expect(spy).toHaveBeenCalledTimes(1);
-    expect(spy).toHaveBeenCalledWith('nested/scopes/admin-page', 'nestedScopesAdminPage');
+    expect(spy).toHaveBeenCalledWith(
+      'nested/scopes/admin-page',
+      'nestedScopesAdminPage',
+    );
+  });
+
+  it('should not alter namespace of the scope', () => {
```
shaharkazaz, line 86:
```suggestion
  it('should keep the original scope name', () => {
```

libs/transloco/src/lib/transloco.service.ts
```diff
@@ -790,7 +790,10 @@ export class TranslocoService implements OnDestroy {
 
   private getMappedScope(scope: string): string {
     const { scopeMapping = {} } = this.config;
-    return scopeMapping[scope] || toCamelCase(scope);
+    return (
+      scopeMapping[scope] ||
+      (this.config.scope.keepNamespaceCasing ? scope : toCamelCase(scope))
```
shaharkazaz, line 795: We are already extracting values from the config in L792

libs/transloco/src/lib/tests/scope-resolver.spec.ts
maartentibau: That was the reason why I was using the `toCamelCase` method to handle this.
I do agree `toCamelCase` is not really the place to do this.

But I could create a new method called

```ts
export function removeInvalidChars(str) {
    return str.replace(/\s+|_|-|\//g, '');
}
```

this way we can also alter `toCamelCase` method to this if you like

```ts
export function toCamelCase(str) {
    return removeInvalidChars(str
        .replace(/(?:^\w|[A-Z]|\b\w)/g, (word, index) => index == 0 ? word.toLowerCase() : word.toUpperCase()));
}
```

libs/transloco/src/lib/transloco.service.ts
maartentibau: Done

libs/transloco/src/lib/tests/scope-resolver.spec.ts
maartentibau: Done

docs/docs/lazy-load/scope-configuration.mdx
maartentibau: Done

docs/docs/getting-started/config-options.mdx
maartentibau: Done

libs/transloco/src/lib/helpers.ts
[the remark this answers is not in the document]
maartentibau: Agreed, changed

libs/transloco/src/lib/tests/scope-resolver.spec.ts
shaharkazaz: Well, underscores for example are valid, just not part of camel-case. the question is how do you see the usage

maartentibau: Hmmm, I think it looked at it the wrong way.

Well as you said in the beginning, it is the responsibility of the consumer to provide a valid `scope` value. 

Of course we could also add a method that would filter out all invalid chars. If you can provide me a list with the allowed `chars` I think I can come up with a regex to only allow these.

shaharkazaz: Let's keep it simple for starters no? it's the consumer's responsibility and if we see that we need to help point to the right direction this kind pf protection can always be added.

maartentibau: Yes, indeed.. we can always add extra protection if need be.

maartentibau:
@shaharkazaz I think we're good to go.
Squashed all the commits

shaharkazaz:
@maartentibau Can you please update from master?

maartentibau:
> @maartentibau Can you please update from master?

Done 🚀

shaharkazaz (approved):

PR #821, opened 2025-01-01
refactor(transloco): tree-shake injection token names in production

arturovt:
In this commit, we tree-shake injection token names in production by inlining the `ngDevMode` condition. This is the same approach that Angular uses.

See, for instance: https://tinyurl.com/4y9dxzsp (long GitHub URL).

`ngDevMode` cannot be extracted into a separate shared variable, as this would result in a separate runtime variable. The condition must always be inlined for the minifier to work correctly.


shaharkazaz (approved):

shaharkazaz:
@arturovt Thanks for the PR! 🥳

PR #822, opened 2025-01-01
fix(transloco): complete subjects when root view is removed

arturovt:


libs/transloco/src/lib/transloco.service.ts
```diff
@@ -685,10 +695,10 @@ export class TranslocoService implements OnDestroy {
   }
 
   ngOnDestroy() {
-    // Caretaker note: since this is the root provider, it'll be destroyed when the `NgModuleRef.destroy()` is run.
-    // Cached values capture `this`, thus leading to a circular reference and preventing the `TranslocoService` from
-    // being GC'd. This would lead to a memory leak when server-side rendering is used since the service is created
-    // and destroyed per each HTTP request, but any service is not getting GC'd.
```
shaharkazaz, line 691: I'd like to keep the comment.
Please leave the following comment above the cache.clear:
```
// As a root provider, this service is destroyed with `NgModuleRef.destroy()`. 
// Cached values retain `this`, causing circular references that block garbage collection, 
// leading to memory leaks during server-side rendering.
```

libs/transloco/src/lib/transloco.service.ts
```diff
@@ -136,7 +145,7 @@ export class TranslocoService implements OnDestroy {
     /**
      * When we have a failure, we want to define the next language that succeeded as the active
      */
-    this.events$.pipe(takeUntilDestroyed()).subscribe((e) => {
```
shaharkazaz, line 139: Please remove the import as it's the only usage in this file

shaharkazaz (commented):
Thanks for the PR 🥳

libs/transloco/src/lib/transloco.service.ts
arturovt: Left the comment.

libs/transloco/src/lib/transloco.service.ts
arturovt: Still pipes the `loadTranslation` observable.

arturovt:
I replaced `ngOnDestroy` with `DestroyRef` to remove the redundant method.

libs/transloco/src/lib/transloco.service.ts
shaharkazaz: Oh right 🤦  haha

shaharkazaz (approved):
🔥

PR #823, opened 2025-01-02
refactor(transloco): tree-shake errors

arturovt:
## Summary by CodeRabbit

* **Bug Fixes**
  * Introduced standardized translation error codes and a formatted error message helper.
  * Added safe dev-mode checks to prevent runtime failures and reduced non-dev console noise.

* **Chores**
  * Updated runtime dev-mode detection and added build annotations for more robust packaging.

* **Tests**
  * Strengthened tests to assert specific error types and explicit dev-mode behaviors.

### DIFF
--- a/libs/transloco-persist-lang/tsconfig.lib.json
+++ b/libs/transloco-persist-lang/tsconfig.lib.json
@@ -6,7 +6,7 @@
     "declaration": true,
     "declarationMap": true,
     "inlineSources": true,
-    "types": [],
+    "types": ["../../@types/angular-globals"],
     "lib": ["dom", "es2018"]
   },
   "exclude": ["src/test-setup.ts", "**/*.spec.ts"],
--- a/libs/transloco/src/lib/tests/default-missing-handler.spec.ts
+++ b/libs/transloco/src/lib/tests/default-missing-handler.spec.ts
@@ -26,7 +26,19 @@ describe('TranslocoMissingHandler', () => {
   it(`GIVEN production mode is enabled
       WHEN handling a missing key
       THEN should not notify a warning message`, () => {
+    // We explicitly bind to the global `ngDevMode` because we no
+    // longer rely on `prodMode`. `ngDevMode` is a build-time variable
+    // defined by Angular CLI since Angular v17, which applies to
+    // all versions supported by Transloco.
+    // Even though Transloco supports Angular v16+, we don't really
+    // care since it has been deprecated for a long time, and this
+    // will be no breaking change for existing apps (since we have a `typeof` check).
+    const originalNgDevMode = ngDevMode;
+    // @ts-expect-error Property 'ngDevMode' does not exist on type 'typeof globalThis'
+    globalThis['ngDevMode'] = false;
     assertLog({ prodMode: true }, false);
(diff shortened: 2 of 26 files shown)


libs/transloco/src/lib/transloco-fallback-strategy.ts
```diff
@@ -1,10 +1,16 @@
 import { Inject, Injectable, InjectionToken } from '@angular/core';
 
 import { TRANSLOCO_CONFIG, TranslocoConfig } from './transloco.config';
+import {
+  formatTranslocoError,
+  TranslocoErrorCode,
+} from './transloco-error-code';
 
 export const TRANSLOCO_FALLBACK_STRATEGY =
-  /* @__PURE__ */ new InjectionToken<TranslocoFallbackStrategy>(
```
shaharkazaz: Why remove the pure comment?

libs/transloco/src/lib/transloco-error-code.ts
```diff
@@ -0,0 +1,12 @@
+export const enum TranslocoErrorCode {
```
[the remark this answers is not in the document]
arturovt: They're not exported publicly.

libs/transloco/src/lib/transloco.config.ts
```diff
@@ -5,6 +5,7 @@ import { AvailableLangs } from './transloco.types';
 export interface TranslocoConfig {
   defaultLang: string;
   reRenderOnLangChange: boolean;
+  /** @deprecated */
```
shaharkazaz: How is this related to this PR? Please keep the changes atomic and relevant.
Furthermore, it's not the same as the ngDevMode, what if I want to see the logs on my STG env which is a production build but internal for my company?

libs/transloco-persist-lang/tsconfig.lib.json
```diff
@@ -6,7 +6,7 @@
     "declaration": true,
     "declarationMap": true,
     "inlineSources": true,
-    "types": [],
+    "types": ["../../@types/angular-globals"],
```
shaharkazaz, line 9: Shouldn't we just define it once at the base and remove the `types: []`?

libs/transloco/src/lib/tests/default-missing-handler.spec.ts
```diff
@@ -26,7 +26,16 @@ describe('TranslocoMissingHandler', () => {
   it(`GIVEN production mode is enabled
       WHEN handling a missing key
       THEN should not notify a warning message`, () => {
+    // We explicitly bind to the global `ngDevMode` because we no
+    // longer rely on `prodMode`. `ngDevMode` is a build-time variable
+    // defined by Angular CLI since Angular v17, which applies to
+    // all versions supported by Transloco.
```
shaharkazaz, line 32: Transloco currently supports v16+

libs/transloco-persist-lang/tsconfig.lib.json
arturovt: it would be cleaner, but it won't work because child `types: []` arrays completely override the base — TypeScript doesn't merge them. Since several libraries intentionally use `types: []` to keep `@types/node` out of Angular builds, the only reliable option is opting in per-library as the branch already does.

shaharkazaz (approved):

PR #866, opened 2025-08-28
fix(transloco): take the last provided scope on selectTranslate

matheo:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/jsverse/transloco/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

Currently we updated the way to provide the scope from:
```ts
{
    provide: TRANSLOCO_SCOPE,
    useValue: 'ourScope',
  }
```
to the util provided by transloco:
```ts
provideTranslocoScope({ scope: 'ourScope', loader })
```
and we started to see untranslated content.

Turns out it was because the util provides `multi: true`, inside our components, we get an array of scopes like:
```ts
[
  {
    scope: 'imported module',
    loader: { ... },
  },
  {
    scope: 'component module',
    loader: { ... },
  },
]
```

## What is the new behavior?

Currently `TranslocoService.selectTranslate()` takes the array of scopes, and only loads the first one.
Turns out, that the scope provided by the module is the last one, so with this PR the `TranslocoService` will load the scope provided last (at the end of the array).

I wonder if all the scopes in the array should be loaded instead.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

@jsverse/transloco v7.6.1

PR #891, opened 2025-11-21
fix(transloco): 🐛 Allow translateSignal API to access any scope

### LINKED ISSUE #875: Bug(transloco): Cannot easily access all TRANSLOCO_SCOPE (multi:true) providers with translateSignal API

- [x] I have searched the existing issues

### Which Transloco package(s) are the source of the bug?

Transloco

### Is this a regression?

No

### Current behavior

Using an app.config.ts with multiple `TRANSLOCO_SCOPE` providers **with inline loaders** I am able to select translations from either scope using the pipe or directive but only have access to the first scope with the `translateSignal` API. I am assuming this is because of the code here which only uses the first scope passed to TranslocoService.selectTranslate: 

https://github.com/jsverse/transloco/blob/master/libs/transloco/src/lib/transloco.service.ts#L337 
`lang = Array.isArray(lang) ? lang[0] : lang;`

Demo component:

```
@Component({
  selector: 'app-root',
  imports: [TranslocoDirective, TranslocoPipe],
  template: `
    <ul *transloco="let t">
      <li>t("a.title"): {{ t('a.title') }}</li>
      <li>t("b.title"): {{ t('b.title') }}</li>
      <li>"a.title" | transloco: {{ 'a.title' | transloco }}</li>
      <li>"b.title" | transloco: {{ 'b.title' | transloco }}</li>
      <li>translateSignal('a.title'): {{ aTitle() }}</li>
      <li>translateSignal('b.title'): {{ bTitle() }}</li>
      <li>translateSignal('title'): {{ title() }}</li>
    </ul>
  `,
})
export class AppComponent {
  public aTitle = translateSignal('a.title');
  public bTitle = translateSignal('b.title');
  public title = translateSignal('title');
}
```

Output:

- t("a.title"): A Title
- t("b.title"): B Title
- "a.title" | transloco: A Title
- "b.title" | transloco: B Title
- translateSignal('a.title'): a.a.title
- translateSignal('b.title'): a.b.title
- translateSignal('title'): A Title

### Expected behavior

I would expect the signal API to work the same way as the directive or pipe since they all have automatic access to the DI system and know what the current scopes are provided to the component.

I want `translateSignal('a.title');` to show "A Title" and `translateSignal('b.title');` to show "B Title" in the UI.

I do realize that I can workaround the situation with code like the below, but I end up having to either import a definition of the inline loader or duplicate it.

```
public explicitBTitle = translateSignal('title', undefined, {
    scope: 'b',
    loader: {
      en_us: () => firstValueFrom(this.http.get<Translation>(`/i18n/b.json`)),
    },
  });
```

### Please provide a link to a minimal reproduction of the bug, if you won't provide a link the issue won't be handled.

https://codesandbox.io/p/devbox/89hkyg

### Transloco Config

```markdown
export const appConfig: ApplicationConfig = {
  providers: [
    provideZoneChangeDetection({ eventCoalescing: true }),
    provideRouter(routes),
    provideHttpClient(),
    provideTransloco({
      config: translocoConfig({
        availableLangs: ['en_us', 'en_ca', 'fr_ca', 'es_us'],
        fallbackLang: 'en_us',
        defaultLang: 'en_us',
        prodMode: true,
        reRenderOnLangChange: true,
        missingHandler: {
          allowEmpty: true,
        },
      }),
    }),
    {
      provide: TRANSLOCO_SCOPE,
      multi: true,
      useFactory: () => {
        const http = inject(HttpClient);

        return {
          scope: 'a',
          loader: {
            en_us: () => firstValueFrom(http.get<Translation>(`/i18n/a.json`)),
          },
        };
      },
    },
    {
      provide: TRANSLOCO_SCOPE,
      multi: true,
      useFactory: () => {
        const http = inject(HttpClient);

        return {
          scope: 'b',
          loader: {
            en_us: () => firstValueFrom(http.get<Translation>(`/i18n/b.json`)),
          },
        };
      },
    },
  ],
};
```

### Please provide the environment you discovered this bug in

```markdown
Transloco: 8.0.2
Angular: 19.2.15
Node: v20.18.3
Package Manager: v10.8.2
OS: 15.6.1 (24G90)
```

### Browser

```markdown
Chrome: Version 140.0.7339.208 (Official Build) (arm64)
```

### Additional context

_No response_

### I would like to make a pull request for this bug

Yes 🚀

maleetz:
Use the autoPrefixKeys=false configuration to find the best provider scope based on the requested key(s), which will then wait for any inline loaders to resolve before trying to translate.

This also includes a small refactor to move the autoPrefixKeys configuration logic into a private function.

✅ Closes: [#875](https://github.com/jsverse/transloco/issues/875)

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: [#875](https://github.com/jsverse/transloco/issues/875)

Using an app.config.ts with multiple TRANSLOCO_SCOPE providers with inline loaders, you can select translations from either scope using the pipe or directive but only have access to the first scope with the translateSignal API.

## What is the new behavior?

To avoid introducing a breaking change but still fix this issue, we can utilize the new autoPrefixKeys configuration property from https://github.com/jsverse/transloco/pull/868. 

If the user has configured their transloco service to not automatically prefix the translation keys with the requested scopes then their translation keys will contain the full path of the translation, including the requested scope. We can utilize that full path to find the correct transloco scope in the providers array to support any inline loaders for that scope.

This PR does not address requesting translations from multiple scopes with inline loaders, if someone needs to do that they should create separate signals.

- [ ] Yes
- [x] No

### DIFF
--- a/libs/transloco/src/lib/transloco.service.ts
+++ b/libs/transloco/src/lib/transloco.service.ts
@@ -38,6 +38,7 @@ import {
   InlineLoader,
   LangDefinition,
   LoadOptions,
+  ProviderScope,
   SetTranslationOptions,
   TranslateObjectParams,
   TranslateParams,
@@ -283,15 +284,11 @@ export class TranslocoService {
 
     if (Array.isArray(key)) {
       return key.map((k) =>
-        this.translate(
-          this.config.scopes.autoPrefixKeys && scope ? `${scope}.${k}` : k,
-          params,
-          resolveLang,
-        ),
+        this.translate(this.autoPrefixKey(k, scope), params, resolveLang),
       ) as any;
     }
 
-    key = this.config.scopes.autoPrefixKeys && scope ? `${scope}.${key}` : key;
+    key = this.autoPrefixKey(key, scope);
 
     const translation = this.getTranslation(resolveLang);
     const value = translation[key];
@@ -322,7 +319,7 @@ export class TranslocoService {
   selectTranslate<T = any>(
     key: TranslateParams,
     params?: HashMap,
-    lang?: string | TranslocoScope | TranslocoScope[],
+    lang?: TranslocoScope | TranslocoScope[],
     _isObject = false,
   ): Observable<T> {
     let inlineLoader: InlineLoader | undefined;
@@ -338,7 +335,10 @@ export class TranslocoService {
       return this.langChanges$.pipe(switchMap((lang) => load(lang)));
     }
 
-    lang = Array.isArray(lang) ? lang[0] : lang;
+    if (Array.isArray(lang)) {
+      lang = this.findBestScope(key, lang);
+    }
+
     if (isScopeObject(lang)) {
       // it's a scope object.
       const providerScope = lang;
@@ -367,6 +367,46 @@ export class TranslocoService {
    */
   private isScopeWithLang(lang: string) {
     return this.isLang(getLangFromScope(lang));
+  }
+
+  private autoPrefixKey(key: string, scope: string | undefined) {
+    return this.config.scopes.autoPrefixKeys && scope ? `${scope}.${key}` : key;
+  }
+
+  // Assume that all requested keys should have the same scope
+  private getRequestedScope(keys: TranslateParams) {
+    const key = Array.isArray(keys) ? keys[0] : keys;
+    return key.split('.')[0];
+  }
+
+  private isScopeObjectArray(
+    translocoScopes: TranslocoScope[],
+  ): translocoScopes is ProviderScope[] {
+    return translocoScopes.every((scope) => isScopeObject(scope));
+  }
+
+  private findBestScope(
+    params: TranslateParams,
+    translocoScopes: TranslocoScope[],
+  ) {
+    // If not using autoPrefixKeys, find the correct scope from the array
+    if (
+      !this.config.scopes.autoPrefixKeys &&
+      this.isScopeObjectArray(translocoScopes)
+    ) {
+      const requestedScope = this.getRequestedScope(params);
+      if (requestedScope) {
+        const matchingProviderScope = translocoScopes.find(
+          (providerScope) =>
+            this.getMappedScope(providerScope.scope) === requestedScope,
+        );
+        if (matchingProviderScope) {
+          return matchingProviderScope;
+        }
+      }
+    }
+
+    return translocoScopes[0];
   }
 
   /**
@@ -399,7 +439,7 @@ export class TranslocoService {
       if (Array.isArray(key)) {
         return key.map((k) =>
           this.translateObject(
-            this.config.scopes.autoPrefixKeys && scope ? `${scope}.${k}` : k,
+            this.autoPrefixKey(k, scope),
             params!,
             resolveLang,
           ),
@@ -407,8 +447,7 @@ export class TranslocoService {
       }
 
       const translation = this.getTranslation(resolveLang);
-      key =
-        this.config.scopes.autoPrefixKeys && scope ? `${scope}.${key}` : key;
+      key = this.autoPrefixKey(key, scope);
 
       const value = unflatten(this.getObjectByKey(translation, key));
       /* If an empty object was returned we want to try and translate the key as a string and not an object */
--- a/libs/transloco/src/lib/tests/service/selectTranslate.spec.ts
+++ b/libs/transloco/src/lib/tests/service/selectTranslate.spec.ts
@@ -1,4 +1,4 @@
-import { fakeAsync } from '@angular/core/testing';
+import { fakeAsync, TestBed } from '@angular/core/testing';
 
 import { createService, runLoader, inlineScope } from '../mocks';
 import { TranslocoService } from '../../transloco.service';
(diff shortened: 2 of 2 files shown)


maleetz:
I also created a branch to showcase an alternative which consolidates the scope loading logic from the directive / pipe so that we can reuse it in selectTranslate for this use case: https://github.com/maleetz/transloco/commit/213423c15d0a789177deae87df3081ed7b7c62b1#diff-c07e4a02be032d92cac1cfbef925cccb4f35c3a4a1697a76a9993f7790ca296c.

Things I like about the consolidation:
  - Less duplicated code
  - More clarity that signals / directive / pipe should operate the same when not auto prefixing keys
  - Removes the confusing lang => path (lang + scope) => back to lang conversion. Instead I just keep a reference to the correct lang in the observable
  - Supports the case of selecting from multiple scopes with one selectTranslate call, but I can't imagine this is a likely use case.

Things I don't like:
  - Riskier change

libs/transloco/src/lib/transloco.service.ts
```diff
@@ -322,7 +319,7 @@ export class TranslocoService {
   selectTranslate<T = any>(
     key: TranslateParams,
     params?: HashMap,
-    lang?: string | TranslocoScope | TranslocoScope[],
+    lang?: TranslocoScope | TranslocoScope[],
```
shaharkazaz, line 322: The string here was to mark that this also accepts a language. I agree that the string is a bit too general, but please revert it for the time being.

shaharkazaz (commented):
Hey @maleetz, thanks for the continued effort on this — really appreciate you sticking with it.
                                                                                                                         
I've merged PR #866 as an immediate fix (takes `[lang.length - 1] instead of [0]`), which addresses the most common case. However, looking at this PR more closely, I think we should go with the alternative approach you yourself explored in [213423c](https://github.com/maleetz/transloco/commit/213423c), your instincts there were right.

The approach you explored is closer, but it still  has the same gap - when `autoPrefixKeys` is on (the default), it doesn't load all scopes. The fix needs to unconditionally forkJoin-load every scope in the array, exactly like the directive does today. The `autoPrefixKeys` setting should only affect how keys are prefixed, not which scopes get loaded.
                                               
To summarize what the implementation should do:                                                                        
  1. When `selectTranslate` receives a scope array, forkJoin-load all scopes (unconditionally, regardless of
  `autoPrefixKeys`)                                                                                                        
  2. Then resolve the translation key against the appropriate scope

The concerns you raised about it are valid but manageable. 
1. The "riskier change" is worth it for correctness. What risks do you foresee? 
2. The extra inline loader wait is a one-time cost per scope, and you already wait for it when working with the pipe/directive.

Let me know if you want to pick this up or if you'd prefer I handle it!

maleetz:
Thanks for the reply! 

As far as I can tell, the auto prefixing functionality only applies to the signal / observable API right now. The directive and pipe do not use auto prefixing as shown here https://github.com/jsverse/transloco/issues/875. 

Demo component:
```
@Component({
  selector: 'app-root',
  imports: [TranslocoDirective, TranslocoPipe],
  template: `
    <ul *transloco="let t">
      <li>t("a.title"): {{ t('a.title') }}</li>
      <li>t("b.title"): {{ t('b.title') }}</li>
      <li>"a.title" | transloco: {{ 'a.title' | transloco }}</li>
      <li>"b.title" | transloco: {{ 'b.title' | transloco }}</li>
      <li>translateSignal('a.title'): {{ aTitle() }}</li>
      <li>translateSignal('b.title'): {{ bTitle() }}</li>
      <li>translateSignal('title'): {{ title() }}</li>
    </ul>
  `,
})
export class AppComponent {
  public aTitle = translateSignal('a.title');
  public bTitle = translateSignal('b.title');
  public title = translateSignal('title');
}
```

Config:
```
    provideTranslocoScope({
      scope: 'a',
      loader: { en: () => import('./i18n/a.json') },
    }),
    provideTranslocoScope({
      scope: 'b',
      loader: { en: () => import('./i18n/b.json') },
    }),
```

Output:

t("a.title"): A Title
t("b.title"): B Title
"a.title" | transloco: A Title
"b.title" | transloco: B Title
translateSignal('a.title'): a.a.title
translateSignal('b.title'): a.b.title
translateSignal('title'): A Title

## This PR
The goal of this PR is to allow the caller to access all scoped translations with inline loaders via the signal / observable API.  Since we can now request the translation without auto prefixing, the only remaining task is to make sure to wait for the required inline loader to load. This can be accomplished by finding the correct loader or, as you found in my previous commit ([213423c](https://github.com/maleetz/transloco/commit/213423c)), waiting for all of them to load.

I chose to go with the "find logic" to keep risk to a minimum since the current selectTranslate code already operated with one scope (even if multiple were provided).

**Feel free to take on the refactor to wait for them all.**

Also, I think this sounds like a good idea (https://github.com/jsverse/transloco/pull/866#issuecomment-4180720523): 

>Update provideTranslocoScope to only use multi: true when multiple scopes are passed in a single call (scopes.length > 1). Single-scope calls will go back to the original non-multi behavior.

Might be a bit challenging for people to adopt though since they would have to make a lot of changes in their apps.

PR #909, opened 2026-04-03
refactor(transloco): replace isBrowser() with ngServerMode for SSR detection

arturovt:
Remove the isBrowser() utility in favor of Angular's native ngServerMode global for server-side rendering detection. This aligns with Angular's recommended approach for SSR guards.

- Replace isBrowser() checks in TranslocoPersistLangService with ngServerMode guards
- Add ngServerMode declaration to @types/angular-globals
- Include angular-globals types in transloco-persist-lang tsconfig.lib.json

---
## Summary by cubic
Switched SSR detection to Angular’s `ngServerMode`, with a fallback to `isPlatformServer`, to avoid running browser APIs on the server and improve tree-shaking.

- **Refactors**
  - Guarded `TranslocoPersistLangService` (constructor, `getCachedLang`, `clear`) with `ngServerMode` or `isPlatformServer(inject(PLATFORM_ID))`; added an `Injector` for runtime `PLATFORM_ID` access.
  - Updated browser utils: `isBrowser()` returns false when `ngServerMode` is true (falls back to `window` check); `getBrowserLang()` drops the direct `isBrowser` check; `getBrowserCultureLang()` short-circuits on `ngServerMode` or non-browser.
  - Declared `ngServerMode` in `@types/angular-globals` and included it in `transloco-persist-lang` `tsconfig.lib.json` types.

<sup>Written for commit c1e2ebcb711ddb18d0fee9fc8d76b5a9f3ae6b20. Summary will update on new commits.</sup>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved server-side rendering compatibility for language detection and persisted language handling.
  - Prevented server-rendered applications from reading, writing, or clearing browser storage.
  - Ensured browser language and culture detection continues to work correctly in client environments.
  - Added support for reliable server-mode detection across Angular build and runtime configurations.
- **Tests**
  - Expanded coverage for browser and server-side language behavior.

### DIFF
--- a/libs/transloco-persist-lang/src/lib/persist-lang.service.ts
+++ b/libs/transloco-persist-lang/src/lib/persist-lang.service.ts
@@ -1,13 +1,19 @@
 import {
+  inject,
+  Injectable,
+  Injector,
+  OnDestroy,
+  PLATFORM_ID,
+} from '@angular/core';
+import { isPlatformServer } from '@angular/common';
+import {
   getBrowserCultureLang,
   getBrowserLang,
-  isBrowser,
   TranslocoService,
 } from '@jsverse/transloco';
 import { isFunction } from '@jsverse/utils';
 import { Subscription } from 'rxjs';
 import { skip } from 'rxjs/operators';
-import { inject, Injectable, OnDestroy } from '@angular/core';
 
 import {
   TRANSLOCO_PERSIST_LANG_CONFIG,
@@ -16,6 +22,7 @@ import {
 
 @Injectable({ providedIn: 'root' })
 export class TranslocoPersistLangService implements OnDestroy {
+  private injector = inject(Injector);
   private service = inject(TranslocoService);
   private storage = inject(TRANSLOCO_PERSIST_LANG_STORAGE);
   private config = inject(TRANSLOCO_PERSIST_LANG_CONFIG);
@@ -24,17 +31,41 @@ export class TranslocoPersistLangService implements OnDestroy {
   private storageKey = this.config.storageKey || 'translocoLang';
 
   constructor() {
-    if (isBrowser()) {
-      this.init();
+    // Prefer `ngServerMode` (Angular v17+) for SSR detection, falling back to `isPlatformServer` for
+    // older Angular versions and MFE scenarios where `ngServerMode` may not be available.
+    // This lets bundlers tree-shake `isPlatformServer` and `PLATFORM_ID` when `ngServerMode` is always defined.
+    if (
+      (typeof ngServerMode !== 'undefined' && ngServerMode) ||
+      isPlatformServer(inject(PLATFORM_ID))
+    ) {
+      return;
     }
+
+    this.init();
   }
 
   getCachedLang(): string | null {
-    return isBrowser() ? this.storage.getItem(this.storageKey) : null;
+    // See SSR guard in the constructor.
+    if (
+      (typeof ngServerMode !== 'undefined' && ngServerMode) ||
+      isPlatformServer(this.injector.get(PLATFORM_ID))
+    ) {
+      return null;
+    } else {
+      return this.storage.getItem(this.storageKey);
+    }
   }
 
   clear() {
-    isBrowser() && this.storage.removeItem(this.storageKey);
+    // See SSR guard in the constructor.
+    if (
+      (typeof ngServerMode !== 'undefined' && ngServerMode) ||
+      isPlatformServer(this.injector.get(PLATFORM_ID))
+    ) {
+      return;
+    }
+
+    this.storage.removeItem(this.storageKey);
   }
 
   private updateStorageOnLangChange(): Subscription {
--- a/libs/transloco-persist-lang/src/lib/persist-lang.service.ts
+++ b/libs/transloco-persist-lang/src/lib/persist-lang.service.ts
@@ -1,13 +1,13 @@
+import { inject, Injectable, OnDestroy, PLATFORM_ID } from '@angular/core';
+import { isPlatformServer } from '@angular/common';
 import {
   getBrowserCultureLang,
   getBrowserLang,
-  isBrowser,
   TranslocoService,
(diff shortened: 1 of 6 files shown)


libs/transloco-persist-lang/src/lib/persist-lang.service.ts
```diff
@@ -24,17 +31,41 @@ export class TranslocoPersistLangService implements OnDestroy {
...
+    }
   }
 
   clear() {
-    isBrowser() && this.storage.removeItem(this.storageKey);
+    // See SSR guard in the constructor.
+    if (
+      (typeof ngServerMode !== 'undefined' && ngServerMode) ||
+      isPlatformServer(this.injector.get(PLATFORM_ID))
+    ) {
```
shaharkazaz, line 64: Let's introduce `#isSSR`

libs/transloco/src/lib/utils/browser.utils.ts
```diff
@@ -1,13 +1,9 @@
...
-  return typeof window !== 'undefined';
-}
-
 /**
  * Returns the language code name from the browser, e.g. "en"
  */
 export function getBrowserLang(): string | undefined {
   let browserLang = getBrowserCultureLang();
-  if (!browserLang || !isBrowser()) {
+  if (!browserLang) {
```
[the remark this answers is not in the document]
shaharkazaz: For consistency I would still use the check here, I wouldn't want to depend on another function that might change

libs/transloco/src/lib/utils/browser.utils.ts
```diff
@@ -26,7 +33,9 @@ export function getBrowserLang(): string | undefined {
  * Returns the culture language code name from the browser, e.g. "en-US"
  */
 export function getBrowserCultureLang(): string {
-  if (!isBrowser()) {
+  // See SSR guard in `isBrowser`. When `ngServerMode` is defined, bundlers can inline
+  // the `isBrowser()` result as `false`, eliminating the runtime call entirely.
+  if ((typeof ngServerMode !== 'undefined' && ngServerMode) || !isBrowser()) {
```
shaharkazaz, line 38: If we just use isBrowser in an SSR env, won't both get shaken out?

libs/transloco-persist-lang/src/lib/persist-lang.service.ts
arturovt: This would not be tree-shakable. Functions being used twice are not inlined.

libs/transloco/src/lib/utils/browser.utils.ts
arturovt: Functions being used twice are not inlined.

shaharkazaz (commented):
Thanks for this — moving to `ngServerMode` for SSR detection is the right direction and the tree-shaking motivation is sound. A few things I'd like to sort out before merging:

**1. Constructor guard changes behavior in non-DOM contexts (main concern)**
Previously `init()` ran only when `isBrowser()` (i.e. `window` exists). Now it runs unless we detect *server* via `ngServerMode`/`isPlatformServer`. In a context that reports `PLATFORM_ID === 'browser'` but has no DOM (web worker, some test setups, SSR misconfig), the old code skipped `init()` while the new code will call `init()` → `storage.getItem(...)`, which can throw if `localStorage` isn't available. The old `window` check was a real safety net here. Is dropping it acceptable, or should the server check retain a `window`/DOM fallback?

**2. Redundant `ngServerMode` term in `getBrowserCultureLang`**
Since `isBrowser()` now already returns `false` when `ngServerMode` is true, the `(ngServerMode) || !isBrowser()` first operand is logically redundant. I understand it's an intentional short-circuit for the minifier — could the comment state that explicitly so it doesn't read as an accidental double-check?

**3. Inconsistency between the two helpers**
`getBrowserLang` still uses a bare `!isBrowser()` while `getBrowserCultureLang` got the explicit `ngServerMode ||` short-circuit. Not a bug, but it undercuts the tree-shaking rationale — could we apply the same pattern to both (or neither)?

**4. `PLATFORM_ID` injection cleanup + document why the guard is inlined**
Two small follow-ups on the SSR guard (the inline duplication itself is intentional — see below):
- Instead of `inject(Injector)` plus a mix of `inject(PLATFORM_ID)` (constructor) and `this.injector.get(PLATFORM_ID)` (methods), store the token once as a field — `private platformId = inject(PLATFORM_ID);` — and reference `this.platformId` in all three guards. That removes the `Injector` dependency and unifies the access pattern. `isPlatformServer` still tree-shakes out of the server bundle (it stays inside the foldable `||`); only the negligible `PLATFORM_ID` token is retained eagerly.
- The guard is deliberately repeated inline rather than extracted into a helper/getter, because a method/getter call is opaque to the minifier: it can't fold `this.isServer` to `true`, so it couldn't eliminate the unreachable `this.init()` chain in the server build. That reasoning isn't documented anywhere (the existing comment only covers the `ngServerMode` fallback) — a one-line note would stop a future reader from "helpfully" DRYing it into a function and silently regressing the tree-shaking.

**5. Stale comment**
`@types/angular-globals/index.d.ts` still says "These declarations are only used during testing" — `ngServerMode` is now used in production library code, so that comment should be updated.

**6. Test coverage**
No tests were added, but this introduces new server/browser branching and the behavioral shift in (1). Could we add coverage for the server path — e.g. set `(globalThis as any).ngServerMode = true` (reset in `afterEach`) to assert `init()` is skipped, `getCachedLang()` returns `null`, and `clear()` is a no-op; plus an `isPlatformServer` case via `{ provide: PLATFORM_ID, useValue: 'server' }`? `browser.utils.ts` currently has no spec at all.

Config-wise the `tsconfig` changes look good — core `transloco` picks up the declaration via `typeRoots`, and scoping persist-lang's `types` to `angular-globals` also properly types the existing `ngDevMode` usage.

medbenmakhlouf (approved):

shaharkazaz (approved):

PR #912, opened 2026-04-06
fix(transloco): avoid duplicate fallback lang request when already cached

### LINKED ISSUE #625: Bug(transloco): transloco loader triggers 2 times on scope translation

- [X] I have searched the existing issues

### Which Transloco package(s) are the source of the bug?

Transloco

### Is this a regression?

No

### Current behavior

When navigating to a page with a scoped translation, it trigger 2 times the loader:

image

`useFallbackTranslation` is set to `true` and both `defaultLang` and `fallbackLang` has the same value.

I also suspect another bug, all translation keys are fine on this scoped component, I don't get why it's loading the fallback lang file as there's no error.

### Expected behavior

If the current lang is the same as the fallbackLang, then it should not load the fallback file as it will already be loaded by default.

If there is no translation error, it should not load the fallback lang file.

### Please provide a link to a minimal reproduction of the bug

https://github.com/ngneat/transloco/tree/master/apps/transloco-playground

### Transloco Config

```markdown
{
  provide: TRANSLOCO_CONFIG,
  useValue: translocoConfig({
    availableLangs: ['fr-FR', 'en-GB'],
    defaultLang: 'fr-FR',
    fallbackLang: 'fr-FR',
    // Remove this option if your application doesn't support changing language in runtime.
    reRenderOnLangChange: true,
    prodMode: !isDevMode(),
    missingHandler: {
      useFallbackTranslation: true,
    },
  }),
},
```

### Please provide the environment you discovered this bug in

```markdown
Transloco: `4.2.2`
Angular: `15.1.5`
Node: `18.13.0`
Package Manager: `pnpm@7.27.0`
Typescript: `4.9.5`
OS: `Windows 11 22H2`
```

### Browser

_No response_

### Additional context

Change `fallbackLang` to `en`, set `useFallbackTranslation` to `true` and navigate to the scoped route:

https://github.com/ngneat/transloco/blob/master/apps/transloco-playground/src/app/transloco-root.module.ts#L45-L48

### I would like to make a pull request for this bug

Yes 🚀

arturovt:
When `useFallbackTranslation` is enabled, loading a secondary language (e.g. 'es') would internally call `getFallbacksLoaders`, which invokes `resolveLoader` directly — bypassing the service cache. This caused a duplicate HTTP request for the fallback lang whenever it had already been fetched (e.g. as the default lang).

Fix: before delegating to `getFallbacksLoaders`, check whether the fallback lang's observable is already in the cache. If it is, reuse it via a mapped `forkJoin` instead of issuing a new network request.

Adds a regression test covering the case where `fallbackLang` matches `defaultLang` and a second language is loaded afterward.

Closes #625 

## PR Checklist

## PR Type

- [x] Bugfix

- [x] No

## Release Notes

* **Bug Fixes**
  * Improved translation loading performance by preventing redundant loader calls when fallback languages are already cached.

* **Tests**
  * Added test coverage for fallback language caching behavior.

### DIFF
--- a/libs/transloco/src/lib/tests/service/missingHandler.spec.ts
+++ b/libs/transloco/src/lib/tests/service/missingHandler.spec.ts
@@ -102,4 +102,34 @@ describe('missingHandler', () => {
       expect(service.translate('empty', { value: 'hello' })).toEqual('');
     }));
   });
+
+  describe('useFallbackTranslation - no duplicate request', () => {
+    it(`GIVEN fallback lang is already cached
+        WHEN loading another lang with useFallbackTranslation
+        THEN should not make a duplicate request for the fallback lang`, fakeAsync(() => {
+      const service = createService({
+        // fallbackLang defaults to 'en' in createService, same as defaultLang
+        missingHandler: {
+          useFallbackTranslation: true,
+        },
+      });
+
+      const loaderSpy = spyOn(
+        (service as any).loader,
+        'getTranslation',
+      ).and.callThrough();
+
+      service.load('en').subscribe();
+      runLoader();
+
+      service.load('es').subscribe();
+      runLoader();
+
+      expect(loaderSpy).toHaveBeenCalledTimes(2);
+      expect(loaderSpy.calls.allArgs()).toEqual([
+        ['en', undefined],
+        ['es', undefined],
+      ]);
+    }));
+  });
 });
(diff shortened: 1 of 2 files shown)


medbenmakhlouf:
@shaharkazaz , will delete this branch as it is closed and also duplicate of https://github.com/jsverse/transloco/pull/999

PR #925, opened 2026-06-01
fix(transloco): memory leak due global service reference not being cleaned up

dscheerens:
This pull request is meant to prevent a memory leak when the `TranslocoService` is destroyed, but the global `service` reference is kept alive.

We're running into this issue with a micro frontend architecture, where the `TranslocoService` is not garbage collected after the Angular application that uses it, is swapped out for another MFE application.

To solve the issue, this PR sets the `service` to `undefined` during the destroy lifecycle phase. As a safety feature this will only be done if the service being destroyed matches the current value held by `service` in case multiple micro frontends are active, where each could be having their own instance of the `TranslocoService`.

---
## Summary by cubic
Fixes a memory leak by clearing the global `service` reference when a `TranslocoService` instance is destroyed. Also adds guards to handle calls when no active service is available, which is common in micro frontend swaps.

- **Bug Fixes**
  - Set global `service` to `undefined` during destroy when it matches the instance, enabling garbage collection.
  - Updated `translate` and `translateObject` to safely handle missing service: log a dev warning and return `''` or `[]`.

<sup>Written for commit 789bfe2baf3caf4873cee42bd86fd12c0ce82e07. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/jsverse/transloco/pull/925?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced the translation service to gracefully handle uninitialized service states by returning safe empty values instead of failing
  * Improved service cleanup to properly clear stale references and prevent memory retention

### DIFF
--- a/libs/transloco/src/lib/transloco.service.ts
+++ b/libs/transloco/src/lib/transloco.service.ts
@@ -72,13 +72,20 @@ import {
   resolveInlineLoader,
 } from './utils/scope.utils';
 
-let service: TranslocoService;
+let service: TranslocoService | undefined;
 
 export function translate<T = string>(
   key: TranslateParams,
   params: HashMap = {},
   lang?: string,
 ): T {
+  if (!service) {
+    if (typeof ngDevMode !== 'undefined' && ngDevMode) {
+      console.warn('No active TranslocoService found');
+    }
+    return '' as T;
+  }
+
   return service.translate<T>(key, params, lang);
 }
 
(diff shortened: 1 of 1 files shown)


shaharkazaz:
@dscheerens
Do you actually use the pure translation functions? If not, you can probably opt out in the next major release. see #881.
I'll clarify in the docs when it's recommended to use those

shaharkazaz:
Feel free to reopen if it's still an issue with the new provide function.

PR #932, opened 2026-06-17
perf(transloco): replace markForCheck with targeted detectChanges in directive

cesco69:
Remove `ChangeDetectorRef.markForCheck()` which marked the entire host component tree as dirty on language change. Attribute strategy now relies on direct DOM manipulation (no CD needed). Structural strategy uses `EmbeddedViewRef.detectChanges()` to check only its own template bindings.

### DIFF
--- a/libs/transloco/src/lib/transloco.directive.ts
+++ b/libs/transloco/src/lib/transloco.directive.ts
@@ -1,5 +1,4 @@
 import {
-  ChangeDetectorRef,
   DestroyRef,
   Directive,
   ElementRef,
@@ -55,7 +54,6 @@ export class TranslocoDirective implements OnInit, OnDestroy, OnChanges {
   private providedLoadingTpl = inject(TRANSLOCO_LOADING_TEMPLATE, {
     optional: true,
   });
-  private cdr = inject(ChangeDetectorRef);
   private host = inject(ElementRef);
   private vcr = inject(ViewContainerRef);
   private renderer = inject(Renderer2);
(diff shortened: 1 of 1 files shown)


shaharkazaz (changes requested):
@cesco69 Thanks for opening the PR!
There is actually a specific reason we used `markForCheck` there. 
`markForCheck` schedules CD, while `detectChanges` runs it synchronously on the view, and given that it's reasonable, you use it all across your app, it's possible it will bloat as `markForCheck` would coalesce all schedules into a single change detection - unlike the `detectChanges` which runs when called.

Consulted with @arturovt and we'd appreciate it if you could provide performance benchmarks that this actually improves performance and not degrades them.
Another thing to notice is that this is actually a breaking change as someone implicitly might relay on that `markForCheck` call

cesco69:
Thanks @shaharkazaz for the thorough review.

You're right! I hadn't considered the coalescing behavior. With N directives, `markForCheck()` batches all the dirty marks into a single CD cycle, while `detectChanges()` would run N separate synchronous checks. That's potentially worse, not better.

The breaking change concern is also valid, consumers might implicitly rely on the host component's CD being triggered by the directive.

I'll close this PR. 

**The "real" optimization would require the translate function to return a signal (so Angular tracks dependencies at the individual binding level).**

Thanks for taking the time to explain the reasoning behind the current approach.

cesco69:
new PR signal based https://github.com/jsverse/transloco/pull/938

PR #945, opened 2026-07-31
fix(package): bump minimum Angular version to 20 and prepare for TypeScript 6.0    

medbenmakhlouf:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

Issue Number: N/A

- The minimum supported `@angular/core` peer version across `transloco`, `transloco-locale`, `transloco-messageformat`, `transloco-persist-lang`, `transloco-persist-translations`, and `transloco-preload-langs` was `>=16.0.0`, and the `rxjs` peer range (`>=6.0.0`) was looser than Angular 20's own requirement (`^6.5.3 || ^7.4.0`).
- `npm audit` reported 65 vulnerabilities, all in dev/build tooling.
- The workspace was not yet validated against TypeScript 6.0, and several files in `transloco-schematics` / `transloco-scoped-libs` relied on implicit `any` types and un-guarded possibly-undefined values that TS 6.0's stricter inference surfaces as real errors.
- In `transloco-schematics/src/scope/index.ts`, `createTranslateFilesFromOptions` was called with a mis-typed `translationsPath` property instead of the expected `translationFilePath`, silently leaving the translation file path `undefined` at runtime for scoped module generation.

### Angular 20 / TypeScript 6.0 preparation

- Bumped `@angular/core` peer dependency to `>=20.0.0` in the 6 affected libraries.
- Synced the `rxjs` peer dependency range to `^6.5.3 || ^7.4.0` (matching Angular 20's own requirement) in `transloco`, `transloco-locale`, and `transloco-messageformat`.
- Applied `npm audit fix` (non-breaking) to reduce vulnerabilities from 65 to 54; remaining findings are all dev-tooling-only and require major/breaking upgrades (tracked separately).
- Fixed the real `translationFilePath` bug in `scope/index.ts` so scoped-module translation files are now correctly generated.
- Added proper type annotations/guards across `ng-migrate.ts`, `scope/index.ts`, `keys-manager/index.ts`, `ngx-migrate/index.ts`, `ngx-migrate/migration-matchers.ts`, `split/index.ts`, and `transloco-scoped-libs` sources, eliminating all implicit-`any` and possibly-undefined issues (no behavior changes apart from the fix above).
- Made `tsconfig.base.json` forward-compatible with TypeScript 6.0 by removing the deprecated `baseUrl` option and switching `paths` to explicit `./`-relative entries (required once `baseUrl` is absent), while keeping `moduleResolution: "node"` so the project remains on TypeScript 5.9.3 for now (blocked externally by `@angular/build`/`ng-packagr` peer ranges capping at `<6.0`).
- Updated TypeScript configuration to use `nodenext` module resolution and added Node types / target updates to `tsconfig.lib.json` / `tsconfig.json` files as needed for library builds.
- Added a `ScopedLib` interface and updated `coerceScopedLibs` for stricter typing in `transloco-scoped-libs`.
- Fixed a typed test expectation in the transloco pipe spec (`pipe-spec`).

### Dependency & tooling alignment

- Updated `tslib` and various package dependencies to their latest compatible versions across the monorepo (multiple passes: general dependency bump, then a follow-up round after resolving conflicts).
- Updated `prettier` to `3.9.6` and downgraded `chalk` back to `4.1.2` (to stay CommonJS-compatible with the tooling that requires it).
- Added a `keys-manager` scope to the changelog config and updated `CONTRIBUTING.md` accordingly.

- [x] Yes
- [ ] No

Consumers currently on Angular < 20 or incompatible `rxjs` versions will need to upgrade to satisfy the new peer dependency ranges. No runtime API changes.

## Other information

- This PR intentionally keeps `typescript` at `5.9.3`; TS 6.0 adoption is prepared but deferred until `@angular/build`/`ng-packagr` lift their `typescript: ">=5.9 <6.0"` peer restriction. Moving to 6.0 later requires only: bumping the `typescript` version and switching `moduleResolution` from `"node"` to `"bundler"` in `tsconfig.base.json` plus the two standalone `tsconfig.schematics.json` files.
- Remaining 54 `npm audit` findings (all dev/build tooling, no runtime impact) are recommended as a separate follow-up major-bump task.

## Summary by CodeRabbit

* **Compatibility**
  * Updated supported Angular, RxJS, and runtime compatibility requirements across packages.
  * Improved compatibility with modern TypeScript and Node.js module resolution.

* **Bug Fixes**
  * Translation transpiler errors now preserve the original underlying cause for clearer diagnostics.
  * Improved handling of missing files, modules, and translation assets during migrations and generation.

* **Refactor**
  * Modernized playground build and development configuration.
  * Strengthened type checking across translation tools and schematics without changing expected behavior.


libs/transloco-optimize/tsconfig.json
```diff
@@ -11,7 +11,9 @@
     }
   ],
   "compilerOptions": {
-    "module": "commonjs",
+    "module": "nodenext",
+    "moduleResolution": "nodenext",
+    "strict": true,
```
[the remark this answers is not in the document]
medbenmakhlouf: Both files already consistently set `"module": "nodenext"` / `"moduleResolution": "nodenext"` (no mismatch between them). I rebuilt the project and `require()`d the emitted output directly — it correctly produces CommonJS (`exports.x = ...`, `require(...)`), which matches the package's `package.json` (`"main": "src/lib/transloco-optimize.js"`, no `"type": "module"`). No regression here.

`tsconfig.lib.json` (the config actually used for compilation, extending `tsconfig.json`) already explicitly sets `"module": "nodenext"` alongside the inherited `"moduleResolution": "nodenext"` — it isn't falling back to the base `esnext`. Ran `tsc -p tsconfig.lib.json --noEmit` with no errors, and the full project build succeeds.

libs/transloco-validator/tsconfig.json
```diff
@@ -11,7 +11,9 @@
     }
   ],
   "compilerOptions": {
-    "module": "commonjs",
+    "module": "nodenext",
+    "moduleResolution": "nodenext",
+    "strict": true,
```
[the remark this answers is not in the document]
medbenmakhlouf: Same story — both already use `nodenext`/`nodenext`, and the built output is verified CommonJS-compatible via a direct `require()` test against `dist/libs/transloco-validator`. No change needed.

libs/schematics-core/eslint.config.mjs
```diff
@@ -0,0 +1,35 @@
+import baseConfig from '../../eslint.config.mjs';
+import jsoncEslintParser from 'jsonc-eslint-parser';
```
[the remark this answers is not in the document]
medbenmakhlouf: done

libs/transloco/src/lib/tests/transpiler.spec.ts
```diff
@@ -167,6 +167,27 @@ describe('TranslocoTranspiler', () => {
...
+
+      expect(() =>
+        transpiler.transpile(getTranspilerParams('[[ upperCase(lowercase) ]]')),
+      ).toThrowError(
+        expect.objectContaining({
+          cause: sentinel,
+        }),
+      );
+
+      spy.mockRestore();
```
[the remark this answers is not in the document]
medbenmakhlouf: done

apps/transloco-playground/project.json
```diff
@@ -59,7 +59,7 @@
       "defaultConfiguration": "production"
     },
     "serve": {
-      "executor": "@nx/angular:dev-server",
+      "executor": "@angular/build:dev-server",
```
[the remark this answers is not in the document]
medbenmakhlouf: I looked into it further. Even in the latest Nx (23.1.1), `@nx/angular:dev-server` unconditionally calls `assertBuilderPackageIsInstalled('@angular-devkit/build-angular')` and dynamically imports it (`dist/src/builders/dev-server/dev-server.impl.js`), regardless of which build executor it's paired with — so upgrading Nx doesn't remove the requirement.

I'm intentionally keeping `@angular/build` (esbuild) as the sole builder here to keep the dependency footprint light. Installing `@angular-devkit/build-angular` just to satisfy `@nx/angular:dev-server` would pull in the entire Webpack toolchain as transitive dependencies (`webpack`, `webpack-dev-server`, `webpack-merge`, `css-loader`, `babel-loader`, `sass-loader`, `@babel/core` + several babel plugins, `terser`, `@ngtools/webpack`, etc.) — none of which would actually be exercised, since the build target (`@nx/angular:application`) is esbuild-based and `@nx/angular:dev-server` would just delegate to Vite anyway in that case.

So this would add real weight and audit surface for zero functional benefit. I'm sticking with `@angular/build:dev-server` paired with `@nx/angular:application` for the build target — confirmed working via `nx serve`/`nx build`. Happy to revisit if we ever need Nx-specific serve features like buildable-library TS path patching.

apps/transloco-playground/project.json
```diff
@@ -7,13 +7,14 @@
...
+      "outputs": ["{options.outputPath.base}"],
       "options": {
-        "outputPath": "dist/apps/transloco-playground",
+        "outputPath": {
+          "base": "dist/apps/transloco-playground"
+        },
         "index": "apps/transloco-playground/src/index.html",
-        "main": "apps/transloco-playground/src/main.ts",
-        "polyfills": "apps/transloco-playground/src/polyfills.ts",
+        "polyfills": ["apps/transloco-playground/src/polyfills.ts"],
```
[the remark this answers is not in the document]
medbenmakhlouf: this is done

shaharkazaz:
@medbenmakhlouf Can you verify the release with the schematics?

medbenmakhlouf:
@shaharkazaz I verified it with  `ng add`  end-to-end and found a real bug. The release builds transloco as ESM ( "type": "module" ), but the schematics compile to CommonJS separately, so Node tries to parse them as ESM and `ng add`  crashes with  exports is not defined. 

Missing piece: a  `{"type": "commonjs"} ` override scoped to the schematics folders (plus  ng-packagr 's auto-generated  .npmignore  was stripping nested package.jsons, so that needed unignoring too). 

Pushed a fix  `ng add`  works now on a fresh app. Our unit tests never caught it since they don't go through real package resolution.

libs/transloco/project.json
```diff
@@ -32,6 +32,9 @@
           },
           {
             "command": "npx tsc -p libs/transloco/tsconfig.schematics.json"
+          },
+          {
+            "command": "printf '\\n!schematics/package.json\\n!schematics-core/package.json\\n' >> dist/libs/transloco/.npmignore"
           }
```
shaharkazaz, line 38: Please wrap in a small ts script

libs/transloco-optimize/package.json
shaharkazaz, line 1: Doesn't this mean it turns into an ESM? flat v6.0.1

libs/transloco-schematics/package.json
shaharkazaz, line 1: schematics need to be commonjs, these are ESM packages no?

libs/transloco-scoped-libs/package.json
shaharkazaz, line 1: same commonjs ESM question

shaharkazaz:
@medbenmakhlouf Let's verify all of the CLI packages as I suspect there might be issues there with CJS → ESM

libs/transloco/project.json
medbenmakhlouf: done

libs/transloco-optimize/package.json
medbenmakhlouf: No, It doesn't.  flat 's  `"type": "module" ` only governs its own files. Our  **transloco-optimize**  dist stays `"type": "commonjs" ` and just does  `require('flat') `. Same as what we did for chalk: bumped Node engine to **>=22** (we're on `>=24`) so native  `require(esm)`  handles pulling in an ESM-only dep so no conversion needed on our side.

libs/transloco-schematics/package.json
medbenmakhlouf: That fix isn't needed for those two. It was only required because  **transloco/schematics**  and  **schematics-core**  are nested inside the main package, whose dist  package.json  gets  `"type": "module"`  from ng-packagr, so the CJS schematics files needed a nested override to avoid being misread as ESM.  **transloco-schematics**  and  **transloco-scoped-libs**  are standalone packages built via tsc, with no  "type"  field, so they already default to CommonJS — nothing to change there.

libs/transloco/package.json
```diff
@@ -25,11 +29,11 @@
...
   "peerDependencies": {
-    "@angular/core": ">=16.0.0",
-    "rxjs": ">=6.0.0"
+    "@angular/core": ">=20.0.0",
+    "rxjs": "^6.5.3 || ^7.4.0"
   },
   "dependencies": {
-    "tslib": "^2.2.0",
+    "tslib": "^2.8.1",
     "@jsverse/utils": "1.0.0-beta.5",
```
shaharkazaz, line 37: TKM uses beta.6

medbenmakhlouf:
Done, I set up a realistic end-to-end test in a fresh project (not our monorepo's node_modules): installed all 4 CLIs from packed tarballs, ran  `keys-manager `extract  on real templates,  `validator`  on the output (plus a duplicate-key fixture to confirm it fails correctly),  `optimize`  to flatten/strip comments (exercises  flat , ESM-only), and  `scoped-libs`  to copy a scoped lib's translations (exercises chalk / chokidar / fs-extra ). All ran clean on Node 24 & Node 22, no `ERR_REQUIRE_ESM`  anywhere.

shaharkazaz (approved):

PR #983, opened 2026-08-09
feat(transloco): v9 deprecation removal, ng update migration and keys-manager Angular 20+ support

shaharkazaz:
## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature
- [x] CI related changes
- [x] Documentation content changes

Three separate gaps ahead of the v9 release:

1. `translocoRead` is still present, marked `@deprecated ... will be removed in Transloco v9`.
2. `@jsverse/transloco` ships no `ng-update` migrations collection at all, so none of the v9 breaking changes can be applied automatically.
3. `@jsverse/transloco-keys-manager` declares `@angular/compiler >= 21.1.0` and `typescript >= 5.9.0`, while the runtime packages accept Angular `>=20`. Putting TKM in an update package group would abort `ng update` for anyone on Angular 20.

Issue Number: N/A

### 1. `fix(keys-manager): support angular 20 and typescript 5.8`

Only one symbol genuinely required Angular 21.1 — `TmplAstSwitchBlockCaseGroup`. `ParenthesizedExpression` has existed since ≤20.0.0, and `@jsverse/angular-utils` peers `>=18.1.0`. The real constraint was two AST **shape** changes in 21.1, so presence checks alone are not enough:

|                  | Angular 20.x – 21.0                       | Angular 21.1+                                   |
| ---------------- | ----------------------------------------- | ----------------------------------------------- |
| `@switch`        | `SwitchBlock.cases`, children on the case | `SwitchBlock.groups`, children on the new group |
| literal map keys | `{ key, quoted }`                         | union tagged with `kind`, to allow spread keys  |

Both are normalized in a new `keys-builder/template/compiler-compat.ts`. It uses a namespace import on purpose: TKM compiles to CJS against an ESM-only `@angular/compiler`, so a named import of a missing symbol is `undefined` and `x instanceof undefined` throws.

On Angular 20 today TKM **crashes** — first at `resolveKeysFromLiteralMap` destructuring an empty filter result, then at the `instanceof`. Verified by stashing the fix, rebuilding, and re-running.

Peers widen to `@angular/compiler >= 20.0.0 < 23.0.0` and `typescript >= 5.8.0 < 7.0.0`. The TypeScript floor is load-bearing: `@angular/compiler-cli@20.0.x` pins `typescript >=5.8 <5.9`, so leaving the peer at `>=5.9` would have re-created the same unsatisfiable-peer abort the compiler widening was meant to remove.

### 2. `feat(transloco): remove deprecated translocoRead input`

Removes the input and its `this.prefix || this.inlineRead` fallback. **Breaking.**

Keys-manager deliberately keeps extracting `translocoRead`, so it still works against codebases that have not migrated.

### 3. `feat(transloco): add ng update migration to v9`

Adds the `ng-update` collection `@jsverse/transloco` never had, plus the packaging to ship it (`exports`, `ng-package.json` assets, `tsconfig.schematics.json`, `.npmignore` overrides). `packageGroup` now covers every published package including keys-manager.

`ng update @jsverse/transloco` will:

- Rename `translocoRead` → `translocoPrefix` in `.html` files and inline `template:` literals, in **both** the attribute form and the `*transloco="…; read: …"` microsyntax. Where an element carries both, the `read` is dropped rather than renamed — `prefix` already won in v8, and renaming would emit a duplicate attribute that fails to parse.
- Add `provideGlobalTranslateFn()` to applications that import `translate`/`translateObject`, restoring the implicit v8 wiring. Libraries are reported instead; projects with no usage are skipped, preserving the SSR/MFE opt-out.
- Report what it cannot fix: the Angular/rxjs floors and the CLI packages' Node `>=22` bump (`ng update` never inspects `engines`).

### 4. `ci(keys-manager): verify angular and typescript ranges`

New `keys-manager-compat` job, wired into the `ci-passed` gate, over four pairs: Angular 20/21.0/21/22 with TypeScript 5.8/5.9/6.0.

Swapping `@angular/compiler` at the workspace root does not work — `@analogjs/vite-plugin-angular` imports `ClassPropertyMapping`, which needs Angular ≥21, so the test tooling breaks before TKM runs. Instead `tools/scripts/verify-keys-manager-compat.mts` packs the built library, installs it into a throwaway project at the target versions and drives the published CLI, asserting all 16 keys still extract. That tests what users install rather than the source tree, and runs locally too.

### 5. `docs: document the v9 breaking changes`

New v9 section in `BREAKING_CHANGES.md`, each entry tied to the commit that introduced it.

- [x] Yes

`translocoRead` is removed. `ng update @jsverse/transloco` handles it.

The other v9 breaking changes (Angular `>=20`, rxjs, Node `>=22`, `provideGlobalTranslateFn()`) landed in earlier PRs; this one documents them and migrates what can be migrated.

## Other information

**Verification.** Full `ci:test` (14 projects), `ci:lint` (15 projects) and `ci:build` (13 projects) pass. Packaging was checked against the built output — `dist/libs/transloco/migrations/**` ships, `migrations/package.json` survives the `.npmignore` pass, and the compiled factory loads. The migration was smoke-tested against the real keys-manager template fixtures.

**One decision left for the release.** `warn-unsupported-options.ts` tells users an unsupported CLI flag *"will throw an error in the next major version"*, but nothing throws yet — the escalation is still a proposal (`libs/transloco-keys-manager/v9.md`, item 3) needing per-command option schemas. `BREAKING_CHANGES.md` records it as **not implemented** rather than claiming behaviour v9 will not have. Either land it before tagging, or the promise rolls to v10.

## Summary by CodeRabbit

- **New Features**
  - Added automated Transloco v9 migrations for template updates and global translation provider configuration.
  - Added migration guidance and compatibility checks for Angular, Node.js, and RxJS requirements.
  - Added Angular 20–22 and TypeScript 5.8–6.x support for Transloco Keys Manager.

- **Breaking Changes**
  - Removed the deprecated `translocoRead` directive input; use `translocoPrefix` instead.
  - Global translation functions now require explicit provider configuration.

- **Bug Fixes**
  - Improved translation-key extraction across supported Angular compiler versions.

### DIFF
--- a/BREAKING_CHANGES.md
+++ b/BREAKING_CHANGES.md
@@ -28,7 +28,7 @@ nx migrate @jsverse/transloco  # Nx
 - The package moved into the main repo and joined the shared version line, so it jumps from `8.1.1` to `9.0.0`.
 - `@angular/compiler` peer dependency is now `>=v20` and `typescript` is now `>=5.8`.
 
-## Transloco Optimize, Scoped Libs, Utils & Validator
+## Transloco Keys Manager, Optimize, Scoped Libs, Utils & Validator
 
 - Node.js `>=22` is now required.
 - Transloco Scoped Libs bumped `chokidar` to v5, which is ESM-only.
--- a/BREAKING_CHANGES.md
+++ b/BREAKING_CHANGES.md
@@ -1,3 +1,38 @@
+# Transloco v9
+
+Most of the changes below are applied automatically:
+
+```bash
+ng update @jsverse/transloco   # Angular CLI
+nx migrate @jsverse/transloco  # Nx
+```
+
+## Transloco
+
+- Removed the deprecated `translocoRead` input, use `translocoPrefix` instead. In the structural form the `read` microsyntax key becomes `prefix`.
+- `translate()` and `translateObject()` now require `provideGlobalTranslateFn()` in the providers. Without it they return `''` / `[]` and warn in dev mode. Omit it in SSR and multi-instance MFE setups.
+- `@angular/core` peer dependency is now `>=v20`.
+- `rxjs` peer dependency is now `^6.5.3 || ^7.4.0`.
+
+## Transloco Locale & Messageformat
+
+- `@angular/core` peer dependency is now `>=v20`.
+- `rxjs` peer dependency is now `^6.5.3 || ^7.4.0`.
+
+## Transloco Persist Lang, Persist Translations & Preload Langs
+
+- `@angular/core` peer dependency is now `>=v20`.
+
+## Transloco Keys Manager
+
+- The package moved into the main repo and joined the shared version line, so it jumps from `8.1.1` to `9.0.0`.
+- `@angular/compiler` peer dependency is now `>=v20` and `typescript` is now `>=5.8`.
+
+## Transloco Optimize, Scoped Libs, Utils & Validator
+
+- Node.js `>=22` is now required.
+- Transloco Scoped Libs bumped `chokidar` to v5, which is ESM-only.
+
 # Transloco v8
 
 ## Transloco
--- a/libs/transloco/migrations/v9/global-translate-fn.ts
+++ b/libs/transloco/migrations/v9/global-translate-fn.ts
@@ -4,37 +4,147 @@ import {
   Tree,
   chain,
 } from '@angular-devkit/schematics';
-import { addRootProvider } from '@schematics/angular/utility/standalone/rules';
-import { getMainFilePath } from '@schematics/angular/utility/standalone/util';
 
+import { loadStandaloneRules, loadTypeScript } from './lazy-deps';
 import { collectFiles, getProjects } from './workspace-utils';
 
-const TRANSLOCO_IMPORT =
-  /import\s+(?:type\s+)?\{([^}]*)\}\s*from\s*['"]@jsverse\/transloco['"]/g;
+const TRANSLOCO_PACKAGE = '@jsverse/transloco';
 
 /** The standalone functions that stopped being auto-wired in v9. */
 const GLOBAL_FNS = ['translate', 'translateObject'];
 
 /**
- * Matches on import specifiers rather than call sites, so a local helper named
- * `translate` isn't mistaken for the global one.
+ * Whether `source` reaches for `translate()`/`translateObject()` at runtime.
+ *
+ * Detection is anchored on the import rather than on call sites, so a local
+ * helper that happens to be named `translate` is not mistaken for the global
+ * one. Reading the import declarations from the AST rather than by pattern is
+ * what lets `import type { translate }` be recognised for what it is: erased
+ * before the app ever runs, so it needs no provider.
  */
 export function usesGlobalTranslateFn(source: string): boolean {
-  TRANSLOCO_IMPORT.lastIndex = 0;
-  let match: RegExpExecArray | null;
+  const ts = loadTypeScript();
+  if (!ts) return false;
 
-  while ((match = TRANSLOCO_IMPORT.exec(source)) !== null) {
-    const imported = match[1].split(',').map((specifier) =>
-      specifier
-        .trim()
-        .split(/\s+as\s+/)[0]
-        .trim(),
+  const file = ts.createSourceFile(
+    'transloco-usage.ts',
+    source,
+    ts.ScriptTarget.Latest,
+    true,
+  );
+
+  for (const statement of file.statements) {
+    if (
+      !ts.isImportDeclaration(statement) ||
+      !ts.isStringLiteral(statement.moduleSpecifier) ||
+      statement.moduleSpecifier.text !== TRANSLOCO_PACKAGE
+    )
+      continue;
+
+    // `import type { ... }` disappears at compile time, so it is not usage.
+    const clause = statement.importClause;
+    if (!clause || clause.isTypeOnly || !clause.namedBindings) continue;
+
+    const bindings = clause.namedBindings;
+
+    // `import * as transloco` only counts if the namespace is actually called.
+    if (ts.isNamespaceImport(bindings)) {
+      if (callsNamespacedGlobalFn(file, bindings.name.text, ts)) return true;
+
+      continue;
+    }
+
+    const imports = bindings.elements.some(
+      (element) =>
+        !element.isTypeOnly &&
+        GLOBAL_FNS.includes((element.propertyName ?? element.name).text),
     );
 
-    if (imported.some((name) => GLOBAL_FNS.includes(name))) return true;
+    if (imports) return true;
   }
 
   return false;
+}
+
+const PROVIDER_FN = 'provideGlobalTranslateFn';
+
+/**
+ * Whether `source` actually calls `provideGlobalTranslateFn()`.
+ *
+ * The check is on a call expression rather than on the text, so a mention in a
+ * comment, a doc string or an unrelated identifier cannot make the migration
+ * believe a project is already wired and skip it. Any call counts, wherever it
+ * lives: extracting providers into their own file and spreading them into
+ * `app.config.ts` is a normal thing to do.
+ */
+export function providesGlobalTranslateFn(source: string): boolean {
+  const ts = loadTypeScript();
+  if (!ts) return false;
+
+  const file = ts.createSourceFile(
+    'transloco-providers.ts',
+    source,
+    ts.ScriptTarget.Latest,
+    true,
+  );
+
+  let found = false;
+
+  const visit = (node: import('typescript').Node): void => {
+    if (found) return;
+
+    if (ts.isCallExpression(node)) {
+      const callee = node.expression;
+      const name = ts.isIdentifier(callee)
+        ? callee.text
+        : ts.isPropertyAccessExpression(callee)
+          ? callee.name.text
+          : undefined;
+
+      if (name === PROVIDER_FN) {
+        found = true;
+
+        return;
+      }
+    }
+
+    ts.forEachChild(node, visit);
+  };
+
+  ts.forEachChild(file, visit);
+
+  return found;
+}
+
+/** Whether `namespace.translate(...)` or `.translateObject(...)` is called. */
+function callsNamespacedGlobalFn(
+  file: import('typescript').SourceFile,
+  namespace: string,
+  ts: NonNullable<ReturnType<typeof loadTypeScript>>,
+): boolean {
+  let found = false;
+
+  const visit = (node: import('typescript').Node): void => {
+    if (found) return;
+
+    if (
+      ts.isCallExpression(node) &&
+      ts.isPropertyAccessExpression(node.expression) &&
+      ts.isIdentifier(node.expression.expression) &&
+      node.expression.expression.text === namespace &&
+      GLOBAL_FNS.includes(node.expression.name.text)
+    ) {
+      found = true;
+
+      return;
+    }
+
+    ts.forEachChild(node, visit);
+  };
+
+  ts.forEachChild(file, visit);
+
+  return found;
 }
 
 /**
@@ -58,6 +168,18 @@ export function addGlobalTranslateFn(): Rule {
       return;
     }
 
+    // Only ships with `@angular/cli` or `@nx/angular`. Without it the provider
+    // cannot be wired, but the template rewrite has already run and stands.
+    const standalone = loadStandaloneRules();
+    if (!standalone) {
+      context.logger.warn(
+        `  ↳ @schematics/angular is not installed, so provideGlobalTranslateFn() could not\n` +
+          `    be added automatically. Add it to your application providers by hand.`,
+      );
+
+      return;
+    }
+
     const rules: Rule[] = [];
     const librariesWithUsage: string[] = [];
     const unresolved: string[] = [];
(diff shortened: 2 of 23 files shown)


BREAKING_CHANGES.md
```diff
@@ -1,3 +1,38 @@
...
+## Transloco Persist Lang, Persist Translations & Preload Langs
+
+- `@angular/core` peer dependency is now `>=v20`.
+
+## Transloco Keys Manager
+
+- The package moved into the main repo and joined the shared version line, so it jumps from `8.1.1` to `9.0.0`.
+- `@angular/compiler` peer dependency is now `>=v20` and `typescript` is now `>=5.8`.
+
+## Transloco Optimize, Scoped Libs, Utils & Validator
```
medbenmakhlouf: @shaharkazaz , also inlclude Keys-Manager in the title.

libs/transloco/migrations/v9/migration-v9.spec.ts
```diff
@@ -0,0 +1,406 @@
...
+      THEN it is left untouched`, () => {
+    expect(migrateTemplate(`<div translocoReadonly="x"></div>`)).toBeNull();
+  });
+
+  it(`GIVEN a template without translocoRead
+      WHEN the template is migrated
+      THEN null is returned so the file is not rewritten`, () => {
+    expect(migrateTemplate(`<div translocoPrefix="a"></div>`)).toBeNull();
+  });
+});
```
medbenmakhlouf, line 381: @shaharkazaz , I think we can strengthen this with additional unit tests. Here are some other cases we could cover:

- `<div *transloco="let t, read: 'a', lang: 'es'"></div>` -> expected:  `<div *transloco="let t, prefix: 'a', lang: 'es'"></div>`
- `<div *transloco="let t, read: 'a', prefix: 'b'"></div>` -> expected: `<div *transloco="let t, prefix: 'b'"></div>`
- `<div *transloco="let t; read: 'a;b'; lang: 'es'"></div>` -> expected:  `<div *transloco="let t; prefix: 'a;b'; lang: 'es'"></div>`
- `<div *transloco="let t; read: 'a;b'; prefix: 'c'"></div>` -> expected:  `<div *transloco="let t; prefix: 'c'"></div>`
- `<div title="foo translocoRead bar"></div>` -> nothing 
- `<my-cmp [label]="translocoRead + suffix"></my-cmp>` -> nothing 
- `<ng-template transloco bind-translocoRead="a">` -> expected:  `<ng-template transloco bind-translocoPrefix="a">` (I doube who can still use these syntax but who knows)
- `<ng-template transloco translocoRead></ng-template>` -> expected:  `<ng-template transloco translocoPrefix></ng-template>`
- `<ng-template transloco translocoRead="a" />` -> expected:  `<ng-template transloco translocoPrefix="a" />`
- `<ng-template transloco [translocoRead]="a" [translocoPrefix]="b"></ng-template>` -> expected: `<ng-template transloco [translocoPrefix]="b"></ng-template>`

libs/transloco/migrations/v9/global-translate-fn.ts
```diff
@@ -0,0 +1,159 @@
+import {
+  Rule,
+  SchematicContext,
+  Tree,
+  chain,
+} from '@angular-devkit/schematics';
+import { addRootProvider } from '@schematics/angular/utility/standalone/rules';
```
medbenmakhlouf: @shaharkazaz `@schematics/angular` is not declared as a dependency under `transloco` lib. However, it may be available through `@angular/cli` or `@nx/angular`. Maybe we could add a check to see if the package is installed before loading it? It could be a nice-to-have improvement, but I don't think it's required.

libs/transloco/migrations/v9/workspace-utils.ts
```diff
@@ -0,0 +1,64 @@
...
+  isApplication: boolean;
+}
+
+/**
+ * `null` when there is no workspace file. `ng update` also runs in repos that
+ * aren't CLI-managed; those still get the template migration, just not the
+ * per-project provider wiring.
+ */
+export function readWorkspace(tree: Tree): WorkspaceSchema | null {
+  const path = ['/angular.json', '/.angular.json'].find((candidate) =>
```
medbenmakhlouf: @shaharkazaz .  in an Nx workspace that uses Angular **without `@nx/angular`** (e.g. `@angular/build` wired up directly), `exists` → `true`, then `read` falls through to the real filesystem, where the file doesn't  exist → throws. The exception escapes `readWorkspace`, which fails the whole rule chain — so users lose the `translocoRead` template rewrite too, not just the provider wiring.

lets wraps the read under `try/catch`

medbenmakhlouf (changes requested):

libs/transloco/migrations/v9/global-translate-fn.ts
shaharkazaz: I think it's ok to assume that this will be available, the project only runs in Angular projects.
We are already depending on it and it's not a gap, in `libs/transloco-schematics/src/split/schema.ts` for example

libs/transloco/migrations/v9/global-translate-fn.ts
```diff
@@ -0,0 +1,99 @@
...
+        .trim()
+        .split(/\s+as\s+/)[0]
+        .trim(),
+    );
+
+    if (imported.some((name) => GLOBAL_FNS.includes(name))) return true;
+  }
+
+  return false;
+}
```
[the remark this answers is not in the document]
shaharkazaz: Agreed, let's also add a spec

libs/transloco/migrations/v9/report-version-floors.ts
```diff
@@ -0,0 +1,69 @@
...
+  '@jsverse/transloco-optimize',
+  '@jsverse/transloco-scoped-libs',
+  '@jsverse/transloco-utils',
+  '@jsverse/transloco-validator',
+];
+
+function readMajor(range: string | undefined): number | null {
+  const match = range?.match(/(\d+)\s*\./);
+
+  return match ? Number(match[1]) : null;
```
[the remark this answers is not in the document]
shaharkazaz: Is there a parser we can use?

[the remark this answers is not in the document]
shaharkazaz: Removed the check rather than fixing the parsing.

Nothing else in the ecosystem does a runtime version check inside a migration — Angular gates each entry with its own `version` key plus peer dependencies, and Nx uses `requires`. So the Angular floor is now declared instead:

```json
"migration-v9": {
  "version": "9.0.0",
  "requires": { "@angular/core": ">=20.0.0" },
  ...
}
```

`ng update` already refuses on an unsatisfiable peer before any migration runs, and Nx evaluates `requires` itself. That leaves no range to parse and avoids adding `semver` as a runtime dependency of the core package just to print a warning.

The Node check stays, and needs no parsing: `process.versions.node` is always a concrete version, and unlike the Angular floor there is no declarative equivalent — `engines.node` is only an `EBADENGINE` warning under npm, and it states intent rather than the interpreter actually running.

libs/transloco/migrations/v9/migration-v9.spec.ts
shaharkazaz: All ten added — and thanks for pushing on this, because three of them turned out to be real bugs rather than missing coverage:

- `<div *transloco="let t, read: 'a', lang: 'es'">` was **never migrated**. `MICRO_READ` was anchored on `(^|;)`, so a `read` key after a comma was skipped entirely — and Angular's microsyntax accepts `,` wherever it accepts `;`.
- `<div title="foo translocoRead bar">` and `<my-cmp [label]="translocoRead + suffix">` were both **rewritten inside the attribute value**. Worse, the same flaw on the removal path silently *deleted* a real binding: `<div data-doc="use translocoPrefix= instead" translocoRead="a">` lost the `translocoRead` entirely, with no warning.
- `<div *transloco="let t; read: 'a;b'; prefix: 'c'">` produced `<div *transloco="let t;b'; prefix: 'c'">`, which fails to parse — I ran the output back through `parseTemplate` to confirm.

Rather than patch the regexes one by one, the migration now **parses for positions** and splices the original text at the reported `keySpan`/`sourceSpan` offsets. Formatting is preserved byte for byte, and the whole class of defects goes away at once — attribute-value false positives can't exist, because the parser never reports value text as an attribute name.

Inline `template:` literals are located through the TypeScript AST for the same reason, restricted to real decorator metadata.

I compared the two implementations directly over these cases: the old one was wrong on **9 of 10**, the new one on **0 of 10**. All ten of your cases are now regression tests, plus the two defects above, plus coverage for `@if`/`@for`/`@switch`/`@defer` blocks and CRLF files.

libs/transloco/migrations/v9/global-translate-fn.ts
shaharkazaz: Coming back on this one — you were right, and I was too quick to wave it off.

Two things changed my mind:

1. The import is **top-level**, and `index.ts` imports the module unconditionally. So a missing `@schematics/angular` doesn't just cost the provider wiring — it takes down the entire migration, template rewrite included.
2. `npm ls` shows it arrives **only** via `@angular/cli` or `@nx/angular`. `@angular/build` has no dependency on it. An Nx workspace wired to `@angular/build` directly, running `nx migrate` (which doesn't require the Angular CLI), is a real gap rather than a theoretical one.

It's now lazily resolved behind a `try`/`catch` in a small `lazy-deps.ts`, together with `@angular/compiler` and `typescript`, which the parser-based rewrite also needs. If any of them is missing the migration degrades to a warning naming what couldn't run, instead of throwing.

libs/transloco/migrations/v9/workspace-utils.ts
shaharkazaz: Done. `readWorkspace` now wraps the whole `exists` → `read` → `JSON.parse` sequence in a single `try`/`catch`, so nothing can escape it and fail the rule chain:

```ts
// `exists` and `read` do not always agree: an Nx host can report a virtual
// `angular.json` as present and then fall through to the real filesystem to
// read it. An exception escaping here would fail the whole rule chain, so
// users would lose the template rewrite as well as the provider wiring.
```

Related, from your other comment: `@schematics/angular` is now lazily loaded behind a `try`/`catch` too, so the provider step degrading no longer costs the template rewrite either.

BREAKING_CHANGES.md
shaharkazaz: Done — the heading is now `Transloco Keys Manager, Optimize, Scoped Libs, Utils & Validator`.

Good catch, and it went further than the docs: keys-manager does declare `engines.node >=22`, but it was also missing from the migration's own `NODE_PACKAGES` list, so the Node warning wouldn't have named it either. Both fixed.

medbenmakhlouf (changes requested):
@shaharkazaz, found two final bugs. Both reproduce with plain code, no special setup.

## 1. `resolveKeysFromLiteralMap` crashes on object spread (Angular 21.1+)

`libs/transloco-keys-manager/src/lib/keys-builder/template/utils.ts:179`

```ts
export function resolveKeysFromLiteralMap(node: LiteralMap): string[] {
  let keys: string[] = [];
  const propertyKeys = node.keys.filter(isLiteralMapPropertyKey);

  for (let i = 0; i < node.values.length; i++) {
    const { key } = propertyKeys[i];
    ...
```

`node.keys` and `node.values` are parallel arrays, but `propertyKeys` filters spread keys out of `keys` before indexing, while `values` still has an entry for every key including spreads. So the loop index desyncs the moment a spread appears.

**Fails on:**
```html
<a [title]="'greeting' | transloco: { ...params, name: 'x' }"></a>
```

**With:**
```
TypeError: Cannot destructure property 'key' of 'propertyKeys[i]' as it is undefined.
    at resolveKeysFromLiteralMap (.../template/utils.js:116:17)
```
Crashes the whole `extract` run, not just a wrong key.

**Fix:** iterate `node.keys` directly and `continue` past non-property keys, so `values[i]` always lines up with the key at the same index:
```ts
for (let i = 0; i < node.keys.length; i++) {
  const mapKey = node.keys[i];
  if (!isLiteralMapPropertyKey(mapKey)) continue;
  const { key } = mapKey;
  const value = node.values[i];
  ...
}
```

## 2. Migration silently drops the read when the prefix sits over a `${}` hole

`libs/transloco/migrations/v9/template-utils.ts` — `classifyPrefix`

Inline templates get their `${...}` interpolation holes masked to spaces before parsing. But if `translocoPrefix`'s *value* is itself one of those holes, `classifyPrefix` sees a masked, blanked-but-non-empty string and classifies it `static` instead of `ambiguous` — so it deletes the `translocoRead` next to it as if the prefix were a known value.

**Fails on:**
```ts
@Component({
  template: `<div translocoPrefix="${p}" translocoRead="fallback"></div>`,
})
```

**Result:** `translocoRead="fallback"` is silently removed, `ambiguous` count stays `0` — no warning, same silent-data-loss shape the `ambiguous` state was built to prevent for bound expressions, just missed for this masked-hole case.

**Fix:** in `classifyPrefix`, check whether the prefix's `sourceSpan` overlaps a masked hole *before* checking `typeof prefix.value === 'string'`, and return `'ambiguous'` if so. Use `sourceSpan`, not `valueSpan` - Angular collapses a whitespace-only value to a zero-length span, so it never overlaps anything.

Let me know if you want me to push the fixes or if you'd rather take them.

After these, this branch is good to go!

medbenmakhlouf (approved):
Well done!

PR #1006, opened 2026-08-29
fix(transloco): type the optional loader constructor param as nullable

arturovt:
`TRANSLOCO_LOADER` is `@Optional()`, so Angular passes `null` when no loader is registered, but the parameter was typed `TranslocoLoader`. A subclass forwarding `inject(TRANSLOCO_LOADER, { optional: true })` to `super()` had to add a non-null assertion to compile.

Widen the parameter to `TranslocoLoader | null`; `this.loader` stays non-null through the existing default-loader fallback.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `TranslocoService` extension so subclasses can be instantiated without explicitly providing a loader.
  * Ensured translation continues to use the default loader when no custom loader is available.

* **Tests**
  * Added coverage verifying optional loader injection and default translation behavior.

### DIFF
--- a/libs/transloco/src/lib/tests/service/extend-service.spec.ts
+++ b/libs/transloco/src/lib/tests/service/extend-service.spec.ts
@@ -1,0 +1,43 @@
+import { inject, Injectable } from '@angular/core';
+import { TestBed } from '@angular/core/testing';
+
+import { TranslocoService } from '../../transloco.service';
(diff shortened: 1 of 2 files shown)


medbenmakhlouf (approved):

medbenmakhlouf:
This is no longer applicable, moved all to `inject` as per this pr #1017

PR #1015, opened 2026-09-08
feat(skills): add AI agent skills for Transloco

medbenmakhlouf:
## What

Adds first-class [Agent Skills](https://skills.sh/) to the repository so AI coding agents can install practical, actionable Transloco guidance with:

```bash
npx skills add https://github.com/jsverse/transloco
```

## Why

Following the approach used by [angular/skills](https://github.com/angular/skills), this gives coding agents version-aware, source-grounded instructions for choosing the right Transloco API, config, scopes/lazy loading, plugins, migrations, and the Keys Manager CLI — instead of relying on stale training data or generic doc dumps.

## Structure

```
skills/
├── README.md                  # architecture overview for maintainers
├── transloco-core/            # translation/template APIs, config, scopes, signals, testing, SSR
├── transloco-plugins/         # locale, messageformat, persist-lang/translations, preload-langs, validator, optimize, scoped-libs
├── transloco-migration/       # ngx-translate / Angular i18n / v8→v9 migration
└── transloco-keys-manager/    # TKM extract/find CLI workflow
```

Each `SKILL.md` is a short, rule-based index (`references/*.md` hold deeper per-topic detail), following the `skills` CLI discovery convention (`skills/<name>/SKILL.md`) — verified locally with `npx skills add <path> --list`, which correctly discovers all 4 skills.

Content is sourced from `docs/v9`, `BREAKING_CHANGES.md`, the v9 migration guide, and the actual library source (directive/pipe/service/providers) — no invented APIs.

## Testing

- Validated all `SKILL.md` frontmatter (`name`/`description` required fields, valid YAML).
- Verified all internal `references/*.md` links resolve.
- Ran `npx skills@latest add <local path> --list` against this branch — confirmed all 4 skills are discovered with correct names/descriptions.

## Review feedback addressed

Applied 16 of 18 CodeRabbit suggestions after verifying each against the actual library source (`transpile()` signature, `InlineLoader` type, `setTranslationKey` options object, testing-module merge behavior, `keysManager` config fields, schematics install step, CLI flag placement, v9-only API list, migration commands, `optimize` programmatic usage, injection-context example, scoped-libs vendor loader, validator CI/lint-staged glob and Node version). Two suggestions were not applied (see PR comment for reasoning): the `read`/`translocoRead` removal note in `template-apis.md` (confirmed correct against source) and the `preload-langs.md` comma/semicolon nit (would break consistency with sibling docs).

## Follow-ups (not in this PR)

- Per-plugin skills could be split out further if maintainers want finer-grained installs.
- Keep `skills/` in sync with `docs/` when the public API changes (see maintenance notes in `skills/README.md`).

## Summary by CodeRabbit

- **Documentation**
  - Added an “AI Agent Skills” section to the README with installation guidance and available skills.
  - Added comprehensive guidance for Transloco core APIs, configuration, scopes, templates, testing, SSR, migrations, plugins, and translation key management.
  - Added plugin references for locale formatting, ICU messages, language and translation persistence, preloading, optimization, scoped libraries, and translation validation.
  - Updated guidance for Transloco v9, Angular compatibility, supported configuration, installation commands, and migration workflows.


skills/transloco-core/references/template-apis.md
```diff
@@ -0,0 +1,102 @@
...
+```html
+<ng-container *transloco="let t">
+  <p>{{ t('title') }}</p>
+  <comp [title]="t('title')"></comp>
+</ng-container>
+```
+
+Inputs:
+
+- `prefix` (was `read`/`translocoRead`; the input and microsyntax key are **removed in v9**, `ng update` renames them) — namespaces all `t()` calls under a key, e.g. `*transloco="let t; prefix: 'dashboard'"` turns `t('title')` into `dashboard.title`.
```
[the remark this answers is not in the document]
medbenmakhlouf: it refers to `read`/`translocoRead` they will be removed not prefix

skills/transloco-plugins/references/scoped-libs.md
```diff
@@ -0,0 +1,58 @@
...
+
+Default strategy copies each library's files individually. Set `"strategy": "join"` in the library's `package.json` `i18n` entry to combine all of a library's files into one `<lang>.vendor.json`, then merge it in a custom loader:
+
+```typescript
+getTranslation(lang: string, { scope }) {
+  const base = this.http.get(`/assets/i18n/${lang}.json`);
+  if (scope) return base;
+  return forkJoin([base, this.http.get(`/assets/i18n/${lang}.vendor.json`)])
+    .pipe(map(([t, vendor]) => ({ ...t, ...vendor })));
+}
```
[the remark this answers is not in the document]
medbenmakhlouf: this is done

skills/transloco-core/references/config.md
```diff
@@ -0,0 +1,47 @@
+# Transloco Configuration
+
+`provideTransloco({ config, loader })` (or `translocoConfig({...})` for the `TRANSLOCO_CONFIG` token directly) accepts:
```
shaharkazaz, line 5: Wouldn't that go stale? why not reference the live docs?

medbenmakhlouf: Good catch! that table can drift. I added a "source of truth" note pointing to the live Config Options docs (jsverse.gitbook.io/transloco/getting-started/config-options) and transloco.config.ts, and added a maintenance rule in skills/README.md requiring option/API summary tables to stay linked to their live docs page and be updated together. Kept the table itself (not just a link) since agents load skills offline and can't reliably fetch live URLs mid-session.

skills/transloco-plugins/references/persist-translations.md
```diff
@@ -0,0 +1,56 @@
...
+| `ttl`        | `number` | Cache time-to-live in seconds.                  |
+| `storageKey` | `string` | Custom key under which translations are stored. |
+
+## Clearing cache
+
+Cache auto-clears on `ttl` expiry, or manually. Inject it in a valid injection context (e.g. a constructor or field initializer) and store it for later use:
+
+```typescript
+export class MyComponent {
+  private persistTranslations = inject(TranslocoPersistTranslations);
```
shaharkazaz: This throws `NullInjectorError`. `TranslocoPersistTranslations` is only `@Injectable()` (not `providedIn: 'root'`), and `provideTranslocoPersistTranslations` registers it solely under the loader token: `{ provide: TRANSLOCO_LOADER, useClass: TranslocoPersistTranslations }` (`libs/transloco-persist-translations/src/lib/transloco-persist-translations.providers.ts:20`). Nothing provides the class token itself. The spec only passes because Spectator provides the class directly.

What works today is `inject(TRANSLOCO_LOADER) as TranslocoPersistTranslations`. The v9 docs page has the same bug.

skills/transloco-plugins/references/persist-translations.md
```diff
@@ -0,0 +1,56 @@
...
+  loader: TranslocoHttpLoader,
+  storage: { useValue: localForage },
+});
+```
+
+## Options
+
+| Option       | Type     | Purpose                                         |
+| ------------ | -------- | ----------------------------------------------- |
+| `ttl`        | `number` | Cache time-to-live in seconds.                  |
```
shaharkazaz: `ttl` is compared against `Date.now()` (`transloco-persist-translations.service.ts:116`: `Date.now() - time >= this.config.ttl`), so it's in **milliseconds**, not seconds.

Note that the default itself is also wrong: `const dayInMilliseconds = 86_400` (`transloco-persist-translations.config.ts:29`) is ~86 seconds, not a day. That's a library bug and I'll handle it separately, but the skill shouldn't say seconds.

skills/transloco-plugins/references/persist-translations.md
```diff
@@ -0,0 +1,56 @@
...
+export class MyComponent {
+  private persistTranslations = inject(TranslocoPersistTranslations);
+
+  clearTranslationsCache() {
+    this.persistTranslations.clearCache();
+  }
+}
+```
+
+Anti-pattern: keeping the default HTTP loader registered alongside this plugin's loader wrapper — that causes duplicate/uncached requests.
```
shaharkazaz: "Duplicate/uncached requests" isn't what happens. `provideTransloco({ loader })` and this plugin both provide `TRANSLOCO_LOADER`, so Angular just keeps whichever was registered last: either the cache is silently bypassed or the plugin wins. There are no duplicate requests. Suggest wording it as "the cache may be silently bypassed".

skills/transloco-plugins/references/optimize.md
```diff
@@ -0,0 +1,44 @@
...
+```typescript
+provideTransloco({
+  config: { flatten: { aot: !isDevMode() } },
+});
+```
+
+## Programmatic usage
+
+```typescript
+import translocoOptimize from '@jsverse/transloco-optimize';
```
shaharkazaz: There's no default export. The package `main` is `src/lib/transloco-optimize.js`, which only exports `getTranslationsFolder(dist)`, `getTranslationFiles(dist)` and `optimizeFiles(paths, commentsKey)` (`libs/transloco-optimize/src/lib/transloco-optimize.ts:23,27,33`). Same in 8.4.0. This snippet won't compile.

skills/transloco-plugins/references/optimize.md
```diff
@@ -0,0 +1,44 @@
...
+npm install -D @jsverse/transloco-optimize
+```
+
+## Wiring it into the build
+
+```json
+// package.json (Angular CLI)
+"scripts": {
+  "transloco:optimize": "transloco-optimize dist/my-app/assets/i18n",
+  "build:prod": "ng build --prod && npm run transloco:optimize"
```
shaharkazaz: `--prod` was removed from the Angular CLI, and the v9 packages require `@angular/core >=20`. It should be plain `ng build` (production is the default configuration).

skills/transloco-plugins/references/scoped-libs.md
```diff
@@ -0,0 +1,62 @@
...
+    base,
+    this.http.get(`/assets/i18n/${realLang}.vendor.json`),
+  ]).pipe(map(([t, vendor]) => ({ ...t, [scope]: vendor[scope] })));
+}
+```
+
+## Webpack
+
+```javascript
+const TranslocoScopedLibsWebpackPlugin = require('@jsverse/transloco-scoped-libs/webpack');
```
shaharkazaz: `@jsverse/transloco-scoped-libs/webpack` doesn't resolve. The published file is `webpack.plugin.js` at the package root and the package has no `exports` map, so it needs `require('@jsverse/transloco-scoped-libs/webpack.plugin')`.

skills/transloco-plugins/references/scoped-libs.md
```diff
@@ -0,0 +1,62 @@
...
+
+This also adds the copied files to `.gitignore` automatically (`--skip-gitignore` to opt out).
+
+## Join strategy
+
+Default strategy copies each library's files individually. Set `"strategy": "join"` in the library's `package.json` `i18n` entry to combine all of a library's files into one `<lang>.vendor.json`, then merge it in a custom loader:
+
+```typescript
+getTranslation(lang: string, { scope }: { scope?: string } = {}) {
+  const base = this.http.get(`/assets/i18n/${lang}.json`);
```
shaharkazaz: For a scoped request this fetches `/assets/i18n/core/en.json`, which doesn't exist under the join strategy. Join only writes `<dist>/<lang>.vendor.json` containing `{ [scope]: {...} }` (`libs/transloco-scoped-libs/src/lib/transloco-scoped-libs.ts:175-177,238-246`).

skills/transloco-plugins/references/scoped-libs.md
```diff
@@ -0,0 +1,62 @@
...
+  if (!scope) return base;
+
+  const [, realLang] = lang.split('/'); // lang arrives as "<scope>/<lang>" for scoped requests
+  return forkJoin([
+    base,
+    this.http.get(`/assets/i18n/${realLang}.vendor.json`),
+  ]).pipe(map(([t, vendor]) => ({ ...t, [scope]: vendor[scope] })));
```
shaharkazaz: This double-prefixes keys. Core already wraps a scoped loader's result as `{ [scope]: translation }` (`libs/transloco/src/lib/transloco.service.ts:637-639`), so returning `{ ...t, [scope]: vendor[scope] }` produces `core.core.title`. The scoped branch should return `vendor[scope]` alone.

skills/transloco-plugins/references/scoped-libs.md
```diff
@@ -0,0 +1,62 @@
...
+## Join strategy
+
+Default strategy copies each library's files individually. Set `"strategy": "join"` in the library's `package.json` `i18n` entry to combine all of a library's files into one `<lang>.vendor.json`, then merge it in a custom loader:
+
+```typescript
+getTranslation(lang: string, { scope }: { scope?: string } = {}) {
+  const base = this.http.get(`/assets/i18n/${lang}.json`);
+  if (!scope) return base;
+
+  const [, realLang] = lang.split('/'); // lang arrives as "<scope>/<lang>" for scoped requests
```
shaharkazaz: `[, realLang] = lang.split('/')` breaks for nested scopes (`a/b/en`). Take the last segment instead.

skills/transloco-plugins/references/scoped-libs.md
```diff
@@ -0,0 +1,62 @@
...
+3. In the library's `package.json`, declare its i18n source:
+   ```json
+   {
+     "name": "@app/core",
+     "i18n": [{ "scope": "core", "path": "src/lib/i18n" }]
+   }
+   ```
+4. In the app's global config (see transloco-core's global config), list the library:
+   ```typescript
+   const config: TranslocoGlobalConfig = { scopedLibs: ['./projects/core/'] };
```
shaharkazaz: The string form only works when `rootTranslationsPath` is set. `transloco-utils` has no default for it, and without it the tool exits with "please specify dist path" (`transloco-scoped-libs.ts:86-94,164`). Worth stating.

skills/transloco-plugins/references/scoped-libs.md
```diff
@@ -0,0 +1,62 @@
...
+   // or, for multiple destinations:
+   // scopedLibs: [{ src: './projects/core', dist: ['./projects/spa/src/assets/i18n'] }]
+   ```
+5. Add a script and run it: `"transloco:extract-scoped-libs": "transloco-scoped-libs --watch"`.
+
+This also adds the copied files to `.gitignore` automatically (`--skip-gitignore` to opt out).
```
shaharkazaz: Only if a `.gitignore` already exists in the cwd (`scoped-libs.utils.ts:27-29`); otherwise nothing is written.

skills/transloco-plugins/references/locale.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Config
+
+```typescript
+provideTranslocoLocale({
+  localeConfig: { global: { date: { dateStyle: 'long', timeStyle: 'long' } } },
+});
+```
+
+Component-level override: `providers: [provideTranslocoLocaleConfig(localeConfig)]`.
```
shaharkazaz: `provideTranslocoLocaleConfig` returns `makeEnvironmentProviders(...)` (`libs/transloco-locale/src/lib/transloco-locale.providers.ts:45-52`). Environment providers aren't allowed in a component's `providers` (it's a type error, and it throws at runtime). `TranslocoLocaleService` is also `providedIn: 'root'`, so it wouldn't see a component-level value anyway.

skills/transloco-plugins/references/locale.md
```diff
@@ -0,0 +1,66 @@
...
+Adds localization (l10n) formatting on top of Transloco's translation (i18n) system, using native `Intl` APIs.
+
+## Setup
+
+```typescript
+// app.config.ts (standalone)
+providers: [provideTranslocoLocale(), ...]
+```
+
+For NgModule apps, provide it (and import `TranslocoLocaleModule` if using the module form) in `TranslocoRootModule`.
```
shaharkazaz: `TranslocoLocaleModule` only declares/exports the four pipes. Importing it in `TranslocoRootModule` doesn't make them available to other modules unless it's re-exported, so the module needs to be imported where the pipes are used.

skills/transloco-plugins/references/locale.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Pipes
+
+```html
+<span>{{ date | translocoDate }}</span>
+<span
```
shaharkazaz: Prettier mangled these snippets (`<span\n>{{ ... }}</span\n>`). `skills/**/*.md` should go in `.prettierignore`, the same way `BREAKING_CHANGES.md` is handled, or the embedded HTML/JSON will keep getting reformatted.

skills/transloco-plugins/references/validator.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Usage — CI (GitHub Actions)
+
+```yaml
+on:
+  pull_request:
+    paths:
+      - 'src/assets/i18n/**.json'
+jobs:
+  validate-translations:
```
shaharkazaz: The job has no `runs-on`, so this workflow is invalid as written. The docs example has `runs-on: ubuntu-latest`.

skills/transloco-plugins/references/validator.md
```diff
@@ -0,0 +1,35 @@
+# Validator (`@jsverse/transloco-validator`)
+
+Lints translation JSON files: validates JSON structure and detects duplicate keys. CLI-only, dev-dependency.
```
shaharkazaz: Minor: it isn't CLI-only. The package `main` default-exports `(paths: string[]) => void`, which throws on the first invalid file (`libs/transloco-validator/src/lib/transloco-validator.ts:5`).

skills/transloco-plugins/references/persist-lang.md
```diff
@@ -0,0 +1,41 @@
...
+  cultureLang,
+  defaultLang,
+}) {
+  return yourLogic;
+}
+
+provideTranslocoPersistLang({ getLangFn, storage: { useValue: localStorage } });
+```
+
+For SSR, use the bundled `cookiesStorage()` instead of `localStorage` (cookies are readable server-side, localStorage is not):
```
shaharkazaz: The plugin never reads the cookie on the server. `cookiesStorage` only touches `document.cookie` (`cookie-storage.ts:14`), and the service does nothing unless `isBrowser()` (`persist-lang.service.ts:27,33`). The real SSR point is only that `useValue: localStorage` throws on the server and `cookiesStorage()` doesn't. Same for "SSR-safe cookie option" on L41.

skills/transloco-plugins/SKILL.md
```diff
@@ -0,0 +1,34 @@
...
+| Eagerly preload languages during idle time                        | `@jsverse/transloco-preload-langs`        | [references/preload-langs.md](references/preload-langs.md)               |
+| Lint translation JSON files (duplicate keys, invalid JSON)        | `@jsverse/transloco-validator`            | [references/validator.md](references/validator.md)                       |
+| Shrink/flatten translation files for production builds            | `@jsverse/transloco-optimize`             | [references/optimize.md](references/optimize.md)                         |
+| Ship translations inside a publishable Angular library (monorepo) | `@jsverse/transloco-scoped-libs`          | [references/scoped-libs.md](references/scoped-libs.md)                   |
+
+All packages follow the same install/version rule as `@jsverse/transloco` itself: match the major version to the installed Angular version (see the **transloco-core** skill's compatibility table), and use the `@next` dist-tag only while targeting Angular >=20 / Transloco v9 (alpha).
```
shaharkazaz: Majors don't match Angular majors: Transloco 9 needs Angular `>=20`, and 5-8 cover 16-19. This should say "pick the major from the compatibility table" rather than "match the major version to the installed Angular version".

skills/transloco-core/SKILL.md
```diff
@@ -0,0 +1,70 @@
...
+- **`translateSignal(key, params?, lang?)`** / **`translateObjectSignal(...)`** (from `@jsverse/transloco`, not the service) — signal-based equivalents of `selectTranslate`/`selectTranslateObject`. Prefer these in **new, signal-based components** instead of subscribing manually.
+- **`translate()`/`translateObject()`** standalone functions — convenience proxies to the service, usable outside DI (e.g. plain functions). Same "must be loaded" caveat as `service.translate()`. **On v9+, require `provideGlobalTranslateFn()`** in root providers, and must be **omitted** in SSR and multi-instance/micro-frontend setups (inject `TranslocoService` directly there instead).
+- **`getTranslation()`/`setTranslation()`/`setTranslationKey()`** — read/write the in-memory translation map directly; use sparingly (e.g. dynamic runtime overrides), not as the primary translation mechanism.
+
+Anti-pattern: calling `translate()` (service method or standalone function) before the language has loaded — it silently returns `''`/warns. If you're not sure translations are loaded yet, use `selectTranslate()`.
```
shaharkazaz: Not what happens. `service.translate()` on a key that isn't loaded yet goes through the missing handler, which **returns the key** (`libs/transloco/src/lib/transloco-missing-handler.ts:31`). In dev it also logs `Missing translation for '<key>'` if `logMissingKey` is on, so it isn't silent either.

The standalone `translate()` returns `''` only when `provideGlobalTranslateFn()` hasn't run (`transloco.service.ts:92-99`), which is a different situation.

skills/transloco-core/SKILL.md
```diff
@@ -0,0 +1,70 @@
...
+
+## 0. Always check the project's installed version first
+
+Angular and Transloco major versions are tightly coupled. Before generating code, inspect `package.json`:
+
+| Angular    | `@jsverse/transloco` | Notes                                                                                                                                                                             |
+| ---------- | -------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `>=20`     | `9.x`                | Alpha/`next` tag while 9.0.0 is unreleased. `translocoRead` input and automatic `translate()`/`translateObject()` wiring are **removed** — requires `provideGlobalTranslateFn()`. |
+| `>=16 <20` | `>=5 <=8`            | Current stable line. `translocoRead` still works (deprecated), `translate()`/`translateObject()` work without extra providers.                                                    |
```
shaharkazaz: 5.x and 6.x were published as `@ngneat/transloco`; the `@jsverse` name starts at 7.0.0. The peer range across 5-8 is `@angular/core >=16.0.0` with no upper bound.

Also, "`translocoRead` still works (deprecated)" across this range isn't accurate: `translocoPrefix` only exists from 7.1.0, which is also where `read` got its deprecation tag.

skills/transloco-core/SKILL.md
```diff
@@ -0,0 +1,70 @@
...
+
+Rules:
+
+- Do not use v9-only APIs (`provideGlobalTranslateFn()`) unless `package.json` shows `@jsverse/transloco` at a stable `>=9.0.0` release, or at a `9.0.0-alpha.x`/`9.0.0-beta.x` prerelease (resolved via the `@next` dist-tag or an explicit pin) — or the user explicitly asks to upgrade.
+- Packages under the old `@ngneat/transloco*` scope are unmaintained — if you see them, recommend migrating to `@jsverse/transloco` (same API, scope rename only) rather than adding new code against them.
+- If no version can be determined (new project), default to the current stable API (v8) unless the user asks for v9/`next`.
```
shaharkazaz: "Default to the v8 API", but this skill recommends `translateSignal` (8.0+), the `activeLang` signal (8.3+ only, missing in 8.0-8.2) and `prefix` (7.1+) without any version floors. An agent on 8.1 would generate `transloco.activeLang()`, which doesn't exist there. Please add the minimum versions.

skills/transloco-core/references/scopes-and-loaders.md
```diff
@@ -0,0 +1,71 @@
...
+@Component({ providers: [provideTranslocoLoadingTpl('<p>Loading...</p>')] })
+```
+
+or inline: `*transloco="let t; loadingTpl: loading"` with `<ng-template #loading>...</ng-template>`. Accepts raw HTML or a component. Can be provided globally in root providers too.
+
+## Extension points ("Hack the Library")
+
+- **`TranslocoLoader`** — override how translation files are fetched (`getTranslation(lang)`).
+- **`TranslocoInterceptor`** — mutate translation data/keys before they're saved (`preSaveTranslation`, `preSaveTranslationKey`).
+- **`TranslocoMissingHandler`** — control what happens when a key is missing (default: warn in dev, return `''`).
```
shaharkazaz: The default missing handler returns the key, not `''`, and the warning is gated on `ngDevMode` + `logMissingKey` (`transloco-missing-handler.ts:24-31`). (The hack-the-library docs page has the same mistake.)

skills/transloco-core/references/scopes-and-loaders.md
```diff
@@ -0,0 +1,71 @@
...
+- **`TranslocoFallbackStrategy`** — control fallback language selection (`getNextLangs(failedLang)`), default uses `config.fallbackLang`.
+
+Register custom implementations with the matching `provideTransloco*` function (`provideTranslocoLoader`, `provideTranslocoInterceptor`, `provideTranslocoMissingHandler`, `provideTranslocoFallbackStrategy`).
+
+## Transpiler
+
+- **`DefaultTranspiler`** — resolves `{{ param }}` interpolation; customize markers via `config.interpolation`.
+- **`FunctionalTranspiler`** — adds `[[ fnName(arg1, arg2) ]]` syntax, resolving `fnName` via DI (`provide: 'fnName', useClass: MyResolver implements TranslocoTranspilerFunction`). Compatible with the default transpiler (opt-in, no migration needed for existing translations). Escape literal commas in args with `\,`.
```
shaharkazaz: Inside a JSON translation file this must be written `\\,`; a bare `\,` is an invalid JSON escape. The runtime splitter expects a single backslash after parsing (`transloco.transpiler.ts:164-166`). The transpiler docs page shows `\\,` correctly.

skills/transloco-core/references/config.md
```diff
@@ -0,0 +1,49 @@
...
+| --------------------------------------- | -------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `availableLangs`                        | —              | List of supported languages (or `{ id, label }[]` objects).                                                                                                       |
+| `defaultLang`                           | `'en'`         | Language used before the active language is resolved.                                                                                                             |
+| `reRenderOnLangChange`                  | `false`        | Set `true` if users can switch language at runtime; keep `false` for static/single-language builds to save memory (renders once, unsubscribes).                   |
+| `prodMode`                              | `false`        | When `true`, disables Transloco's console warnings. Wire to `!isDevMode()`.                                                                                       |
```
shaharkazaz: `prodMode` isn't read anywhere in the libraries. It only appears in `transloco.config.ts:8,38` and `transloco-testing.module.ts:80`. Every warning is gated on `ngDevMode` instead. It's effectively dead config, so the skill shouldn't tell agents to wire it up expecting an effect.

skills/transloco-core/references/config.md
```diff
@@ -0,0 +1,49 @@
...
+| `fallbackLang`                          | —              | One or more languages to fall back to when a translation/file is missing. Customize selection via a `TranslocoFallbackStrategy` (see hack-the-library reference). |
+| `failedRetries`                         | `2`            | Retry count for failed translation file loads.                                                                                                                    |
+| `missingHandler.allowEmpty`             | `false`        | Allow empty string values for missing keys instead of falling back.                                                                                               |
```
shaharkazaz: `allowEmpty` applies only to keys that exist with the value `''` (`transloco.service.ts:701`: `allowEmpty && value === ''`). Truly missing keys still go to the fallback or the handler.

skills/transloco-core/references/config.md
```diff
@@ -0,0 +1,49 @@
...
+| `availableLangs`                        | —              | List of supported languages (or `{ id, label }[]` objects).                                                                                                       |
+| `defaultLang`                           | `'en'`         | Language used before the active language is resolved.                                                                                                             |
+| `reRenderOnLangChange`                  | `false`        | Set `true` if users can switch language at runtime; keep `false` for static/single-language builds to save memory (renders once, unsubscribes).                   |
+| `prodMode`                              | `false`        | When `true`, disables Transloco's console warnings. Wire to `!isDevMode()`.                                                                                       |
+| `fallbackLang`                          | —              | One or more languages to fall back to when a translation/file is missing. Customize selection via a `TranslocoFallbackStrategy` (see hack-the-library reference). |
```
shaharkazaz: There's no "hack-the-library reference" in this skill; that content lives in `scopes-and-loaders.md` under "Extension points". Also, the defaults shown as `—` are actually `[]` for `availableLangs`/`fallbackLang` and `false` for `flatten.aot` (`transloco.config.ts:40-48`).

skills/transloco-core/references/config.md
```diff
@@ -0,0 +1,49 @@
...
+| `missingHandler.useFallbackTranslation` | `false`        | Use `fallbackLang`'s value for a key missing in the active language.                                                                                              |
+| `missingHandler.logMissingKey`          | `true`         | Log missing-key warnings in dev mode.                                                                                                                             |
+| `flatten.aot`                           | —              | Enables AOT flattening compatibility; use together with `@jsverse/transloco-optimize`.                                                                            |
+| `interpolation`                         | `['{{', '}}']` | Start/end markers for params in translation strings, e.g. `Hello {{name}}`.                                                                                       |
+| `scopes.keepCasing`                     | `false`        | Preserve a scope's original casing as its namespace instead of camel-casing it (ignored if an `alias` is set).                                                    |
```
shaharkazaz: `scopes.autoPrefixKeys` (default `true`, `transloco.config.ts:53`) is missing from the table.

On the new source-of-truth link: it helps, but the rows above already disagree with the source (`prodMode`, `allowEmpty`, defaults), so the link alone doesn't stop drift. I'd rather drop the copied table and keep only the decision rules.

skills/transloco-core/references/template-apis.md
```diff
@@ -0,0 +1,102 @@
...
+Translation:
+
+- `translate(key, params?, lang?)` — sync; requires translations already loaded.
+- `selectTranslate(key, params?, lang?)` — `Observable<string>`; auto-loads, updates on language change. Default choice for anything rendered reactively.
+- `translateObject(key, params?)` / `selectTranslateObject(key, params?)` — same pair, but for a nested object/array of translations.
+- `getTranslation()` / `getTranslation(lang)` / `getTranslation('scope/lang')` — sync full translation map.
+- `selectTranslation(lang?)` — `Observable` of the full translation map, re-emits on language change.
+- `setTranslation(translation, lang?, { merge })` — manually set/merge translation data.
+- `setTranslationKey(key, value, { lang?, emitChange? })` — set a single key.
+- `events$` — observable of `translationLoadSuccess` / `langChanged` events (only fires on load from source, not from cache).
```
shaharkazaz: Incomplete and partly wrong: `translationLoadFailure` is also emitted (`transloco.types.ts:12-22`). "Not from cache" only applies to load events. `langChanged` fires on every `setActiveLang`, including via `setTranslation`/`setTranslationKey` with the default `emitChange: true` (`transloco.service.ts:246,657`).

skills/transloco-core/references/template-apis.md
```diff
@@ -0,0 +1,102 @@
...
+
+Language:
+
+- `getDefaultLang()` / `setDefaultLang(lang)`
+- `getActiveLang()` / `setActiveLang(lang)` — prefer the `activeLang` signal (below) in signal-based code.
+- `getAvailableLangs()` / `setAvailableLangs(langs)`
+- `setFallbackLangForMissingTranslation({ fallbackLang })` — single language only, even if an array is passed only the first is used.
+- `langChanges$` — emits immediately on subscribe with the current language, then on every change.
```
shaharkazaz: It also re-emits the same language after `setTranslation`, because there's deliberately no `distinctUntilChanged` (`transloco.service.ts:203-205,657`).

skills/transloco-core/references/template-apis.md
```diff
@@ -0,0 +1,102 @@
...
+- `setTranslation(translation, lang?, { merge })` — manually set/merge translation data.
+- `setTranslationKey(key, value, { lang?, emitChange? })` — set a single key.
+- `events$` — observable of `translationLoadSuccess` / `langChanged` events (only fires on load from source, not from cache).
+
+Language:
+
+- `getDefaultLang()` / `setDefaultLang(lang)`
+- `getActiveLang()` / `setActiveLang(lang)` — prefer the `activeLang` signal (below) in signal-based code.
+- `getAvailableLangs()` / `setAvailableLangs(langs)`
+- `setFallbackLangForMissingTranslation({ fallbackLang })` — single language only, even if an array is passed only the first is used.
```
shaharkazaz: Worth noting it's a no-op unless `missingHandler.useFallbackTranslation` is `true` (`transloco.service.ts:692,813-817`).

skills/transloco-core/references/template-apis.md
```diff
@@ -0,0 +1,102 @@
...
+Good for simple static text nodes that don't need the `let t` context function.
+
+## `TranslocoService` API (inject via `inject(TranslocoService)`)
+
+Translation:
+
+- `translate(key, params?, lang?)` — sync; requires translations already loaded.
+- `selectTranslate(key, params?, lang?)` — `Observable<string>`; auto-loads, updates on language change. Default choice for anything rendered reactively.
+- `translateObject(key, params?)` / `selectTranslateObject(key, params?)` — same pair, but for a nested object/array of translations.
+- `getTranslation()` / `getTranslation(lang)` / `getTranslation('scope/lang')` — sync full translation map.
```
shaharkazaz: With no argument, `getTranslation()` returns the whole `Map<lang, Translation>` of every loaded language, not the active language's map (`transloco.service.ts:563-578`).

skills/transloco-core/references/testing-and-ssr.md
```diff
@@ -0,0 +1,57 @@
...
+  translocoConfig: { availableLangs: ['en', 'es'], defaultLang: 'en' },
+  preloadLangs: true,
+});
+```
+
+Anti-pattern: mocking `TranslocoService` by hand in every spec — use `TranslocoTestingModule` so real directive/pipe behavior (memoization, scope resolution) is exercised.
+
+## SSR
+
+- Run `ng add @jsverse/transloco@next` (or the SSR option in `ng-add`) to get SSR-specific wiring: it switches the loader's base path from relative to **absolute**, since relative paths don't resolve on the server.
```
shaharkazaz: The SSR wiring comes from choosing the `ssr` option in ng-add (`schematics/ng-add/schema.json`, `index.ts:252`); `@next` has nothing to do with it. The loader gets prefixed with `${environment.baseUrl}` (`generators/http-loader.gen.ts:18`). This was copied from the SSR docs page, which is itself out of date.

skills/transloco-core/references/testing-and-ssr.md
```diff
@@ -0,0 +1,57 @@
...
+- Transloco adds a `baseUrl` key to `environment.ts` — set it to your app's origin per environment.
```
shaharkazaz: `baseUrl` is only added to files under `src/environments/`, and only if that folder exists (`schematics-core/utils/workspace.ts:37-50`). Current Angular CLI projects don't create it. (Separately, the generated loader references `environment` without importing it; I'll fix that on our side.)

skills/transloco-core/references/testing-and-ssr.md
```diff
@@ -0,0 +1,57 @@
...
+- **Do not use `provideGlobalTranslateFn()`** (the provider enabling the standalone `translate()`/`translateObject()` functions) in SSR apps, nor in multi-instance/micro-frontend setups — both run more than one Transloco instance and a global function can't know which one to resolve against. Inject `TranslocoService` directly instead.
```
shaharkazaz: That's not the actual reason. Per the provider's JSDoc, the module-level reference persists across requests and leaks state between users (`transloco.providers.ts:56-57`), and the provider is a no-op on the server (`:70`), so the standalone `translate()` returns `''` during SSR.

skills/transloco-keys-manager/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Key CLI options (`extract`)
+
+| Flag                      | Purpose                                                                                           | Default                                       |
+| ------------------------- | ------------------------------------------------------------------------------------------------- | --------------------------------------------- |
+| `-i, --input`             | Source dirs to scan                                                                               | `${sourceRoot}/app` (or `/lib` for libraries) |
+| `-o, --output`            | Output dir for translation files                                                                  | `${sourceRoot}/assets/i18n`                   |
+| `-l, --langs`             | Languages to generate                                                                             | `[en]`                                        |
+| `-f, --fileFormat`        | `json` or `pot`                                                                                   | `json`                                        |
```
shaharkazaz: `--fileFormat` doesn't exist: the flag is `--file-format` (`libs/transloco-keys-manager/src/lib/cli-options.ts:36`). `camelCase: true` only affects the parsed output keys, and passing `--fileFormat` throws `UNKNOWN_OPTION`.

skills/transloco-keys-manager/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+| `-m, --marker`            | Marker function name                                                                              | `t`                                           |
+| `-r, --replace`           | Overwrite existing files instead of merging                                                       | `false`                                       |
+| `-R, --remove-extra-keys` | Remove keys no longer used in code                                                                | `false`                                       |
+| `-d, --defaultValue`      | Placeholder value template (supports `{{key}}`, `{{keyWithoutScope}}`, `{{scope}}`, `{{params}}`) | `Missing value for {{key}}`                   |
```
shaharkazaz: Should be `--default-value` (`cli-options.ts:74`). Also, the default isn't a template: the config default is `undefined` (`config.ts:57`) and the output is the literal `Missing value for '<key>'` (`add-key.ts:29-30`). Placeholders are replaced once each (`.replace`, not `replaceAll`).

skills/transloco-keys-manager/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+| `-u, --unflat`            | Nest keys instead of flat dotted keys                                                             | flat                                          |
+| `--sort`                  | Sort keys                                                                                         | `false`                                       |
+| `-c, --config`            | Root dir to search for `transloco.config.ts`                                                      | `process.cwd()`                               |
+| `--project`               | Target Nx/Angular project (prefixes defaults with its `sourceRoot`)                               | `defaultProject`                              |
+
+`find` also has `-a, --add-missing-keys` (add keys `find` flagged as missing), `-e, --emit-error-on-extra-keys` (fail CI on unused keys), and `-p, --translationsPath`.
```
shaharkazaz: Should be `--translations-path` (`cli-options.ts:94`).

skills/transloco-keys-manager/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+| ------------------------- | ------------------------------------------------------------------------------------------------- | --------------------------------------------- |
+| `-i, --input`             | Source dirs to scan                                                                               | `${sourceRoot}/app` (or `/lib` for libraries) |
+| `-o, --output`            | Output dir for translation files                                                                  | `${sourceRoot}/assets/i18n`                   |
+| `-l, --langs`             | Languages to generate                                                                             | `[en]`                                        |
+| `-f, --fileFormat`        | `json` or `pot`                                                                                   | `json`                                        |
+| `-m, --marker`            | Marker function name                                                                              | `t`                                           |
```
shaharkazaz: `-m/--marker` is the **comment** marker (`t(...)` in extraction comments; `cli-options.ts:46`, `add-comment-section-keys.ts:42`), not the `marker()` function name. The function is detected by its import and alias. As written, readers will confuse the two.

skills/transloco-keys-manager/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+| `-r, --replace`           | Overwrite existing files instead of merging                                                       | `false`                                       |
+| `-R, --remove-extra-keys` | Remove keys no longer used in code                                                                | `false`                                       |
+| `-d, --defaultValue`      | Placeholder value template (supports `{{key}}`, `{{keyWithoutScope}}`, `{{scope}}`, `{{params}}`) | `Missing value for {{key}}`                   |
+| `-u, --unflat`            | Nest keys instead of flat dotted keys                                                             | flat                                          |
+| `--sort`                  | Sort keys                                                                                         | `false`                                       |
```
shaharkazaz: Missing the `-s` alias (`cli-options.ts:62-66`).

skills/transloco-keys-manager/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+| `-c, --config`            | Root dir to search for `transloco.config.ts`                                                      | `process.cwd()`                               |
```
shaharkazaz: The default search root is the project's `sourceRoot`, not `process.cwd()` (`resolve-config.ts:25`: `inlineConfig.config || sourceRoot`).

skills/transloco-keys-manager/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+| `--project`               | Target Nx/Angular project (prefixes defaults with its `sourceRoot`)                               | `defaultProject`                              |
+
+`find` also has `-a, --add-missing-keys` (add keys `find` flagged as missing), `-e, --emit-error-on-extra-keys` (fail CI on unused keys), and `-p, --translationsPath`.
+
+Only a subset of these can be set once via the shared `keysManager` block in the project's `transloco.config.ts` (see transloco-core's global config reference): `input`, `output`, `fileFormat`, `marker`, `addMissingKeys`, `emitErrorOnExtraKeys`, `replace`, `defaultValue`, `unflat`, `sort` — plus the top-level `langs` and `rootTranslationsPath` fields. CLI-only options like `-c/--config` and `--project`, and the unsupported `removeExtraKeys`, cannot be configured there and must be passed as CLI flags.
```
shaharkazaz: `removeExtraKeys` works in `transloco.config`; it just isn't in the type (`transloco-utils.types.ts:7-18`). `resolve-config.ts:75` spreads the whole `keysManager` object, so it's applied, and at most you get a TS error. "Unsupported" is too strong.

skills/transloco-keys-manager/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+- The `prefix` input — `t('title')` under `*transloco="let t; prefix: 'dashboard'"` extracts as `dashboard.title`.
+- Literal params — `t('key', {value: '123'})`; **dynamic param objects are not extracted** (`translocoService.translate('key', this.myParams)` won't be picked up).
+- Static ternaries in pipe/directive/attribute usage (`condition ? 'keyOne' : 'keyTwo' | transloco`).
+
+### Dynamic keys
+
+For computed keys the extractor can't statically resolve, add an extraction comment:
+
+```typescript
+// TypeScript — file must import from (or comment-reference) '@jsverse/transloco'
```
shaharkazaz: Not required: comment markers are extracted even when "transloco" appears nowhere in the file (`keys-builder/typescript/index.ts:56-63`). The import only gates the service/function/signal extractors (`:38-40`). The docs hint this came from is outdated.

skills/transloco-keys-manager/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+```bash
+npm run i18n:extract
+```
+
+Scans source for translation key usage and writes/updates the target translation JSON. It understands, out of the box:
+
+- Scopes declared via `provideTranslocoScope(...)` — extracts into the matching subfolder (e.g. `i18n/admin/en.json`).
+- Inline loaders — map them via `scopePathMap` in `transloco.config.ts` (see transloco-core's global config reference) so keys land in the right folder.
+- The `prefix` input — `t('title')` under `*transloco="let t; prefix: 'dashboard'"` extracts as `dashboard.title`.
+- Literal params — `t('key', {value: '123'})`; **dynamic param objects are not extracted** (`translocoService.translate('key', this.myParams)` won't be picked up).
```
shaharkazaz: The key **is** extracted with dynamic params; only the param names are skipped (`build-keys-from-ast-nodes.ts:33-46`), and they only matter through `{{params}}` in `defaultValue`. Same for the anti-pattern on L120.

skills/transloco-keys-manager/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+description: Automates Transloco translation key workflows using the Transloco Keys Manager (TKM) CLI - extracting keys used in code/templates into translation files, finding missing or unused (extra) keys across languages, and marking dynamic/computed keys for extraction. Trigger when the user wants to auto-generate translation JSON from source code, find missing or unused translation keys, or configure the extract/find CLI commands.
+license: MIT
+metadata:
+  author: Transloco maintainers
+  version: '1.0'
+---
+
+# Transloco Keys Manager (TKM)
+
+CLI tool (`@jsverse/transloco-keys-manager`) that scans TypeScript/HTML for translation key usage and keeps translation JSON files in sync — use it instead of manually adding/removing keys across every language file. Requires Node.js `>=22`, `@angular/compiler >=20`, `typescript >=5.8` (v9+; as of v9 it lives in the main `jsverse/transloco` monorepo but its CLI/config are unchanged from v8).
```
shaharkazaz: Missing the upper bounds: `@angular/compiler >= 20.0.0 < 23.0.0`, `typescript >= 5.8.0 < 7.0.0` (`libs/transloco-keys-manager/package.json`).

skills/transloco-keys-manager/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+```
+
+```json
+"scripts": {
+  "i18n:extract": "transloco-keys-manager extract",
+  "i18n:find": "transloco-keys-manager find"
+}
+```
+
+Or via schematics: install `@jsverse/transloco-schematics` as a dev dependency, then run `ng g @jsverse/transloco-schematics:keys-manager` (choose CLI, Webpack plugin, or both). The schematic then installs `@jsverse/transloco-keys-manager` itself, unpinned — it always resolves the `latest` dist-tag regardless of which channel the schematic came from.
```
shaharkazaz: The schematic does more than this: it always installs `ngx-build-plus` (even for "CLI"), switches the serve builder to `ngx-build-plus:dev-server` for Webpack/Both, and errors out without `rootTranslationsPath`/`langs` unless you pass flags (`libs/transloco-schematics/src/keys-manager/index.ts:22-40,88-106`). Worth mentioning, since an agent will run it blind.

skills/transloco-migration/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+| -------------------------------------------------------- | ------------------------------------------------------- |
+| `'key' \| translate`                                     | `'key' \| transloco`                                    |
+| `'key' \| translate:{name:'x'}`                          | `'key' \| transloco:{name:'x'}`                         |
+| `[translate]="'HELLO'" [translateParams]="{...}"`        | `[transloco]="'HELLO'" [translocoParams]="{...}"`       |
+| `import { TranslateService } from '@ngx-translate/core'` | `import { TranslocoService } from '@jsverse/transloco'` |
+| `translateService.currentLang`                           | `translateService.getActiveLang()`                      |
+| `translateService.onLangChange`                          | `translateService.langChanges$`                         |
+| `translateService.use(...)`                              | `translateService.setActiveLang(...)`                   |
+| `translateService.instant(...)`                          | `translateService.translate(...)`                       |
+| `translateService.get(...)`                              | `translateService.selectTranslate(...).pipe(take(1))`   |
```
shaharkazaz: ngx-migrate doesn't add `take(1)`. It maps `get` to `selectTranslate` only (`libs/transloco-schematics/src/ngx-migrate/migration-matchers.ts:93`), and the schematic's own doc says `take(1)` needs to be added manually.

skills/transloco-migration/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+# Nx: nx g @jsverse/transloco-schematics:ngx-migrate
+```
+
+It recursively rewrites HTML/TS files. Key replacements it performs (validate matches when reviewing/writing manual patches too):
+
+| ngx-translate                                            | Transloco                                               |
+| -------------------------------------------------------- | ------------------------------------------------------- |
+| `'key' \| translate`                                     | `'key' \| transloco`                                    |
+| `'key' \| translate:{name:'x'}`                          | `'key' \| transloco:{name:'x'}`                         |
+| `[translate]="'HELLO'" [translateParams]="{...}"`        | `[transloco]="'HELLO'" [translocoParams]="{...}"`       |
```
shaharkazaz: This rewrite doesn't happen. The directive regex is `(translate|\[translate(?:Params)?\])=("|')[^"']*\2` (`migration-matchers.ts:21`), which stops at the inner quote, so `[translate]="'HELLO'"` and `[translateParams]="{value: 'world'}"` stay unchanged. Only `translate="HELLO"` and `[translate]="key"` are rewritten.

skills/transloco-migration/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+
+v9 is currently alpha under the npm `next` tag (`@jsverse/transloco@next`). Confirm the user actually wants to move to alpha/v9 before applying this — otherwise stay on the current v8.x stable line.
+
+```bash
+ng update @jsverse/transloco --next   # or: nx migrate @jsverse/transloco@next && nx migrate --run-migrations
+```
+
+The migration automates:
+
+- Renaming the removed `translocoRead` input / `read` microsyntax key to `translocoPrefix` / `prefix` (in both external templates and inline `template` strings). It **skips and logs** templates that fail to parse, and any element setting both `prefix` and `translocoRead` where the prefix is a dynamic/interpolated expression (only the running app can tell whether it evaluates empty) — review those by hand.
```
shaharkazaz: In the dynamic-prefix + `translocoRead` case the migration **drops the read and warns**; it doesn't skip the element (`libs/transloco/migrations/v9/template-utils.ts:331-335`, `index.ts:47-54`). Only unparsable templates are skipped. The migration guide says the same ("The read is dropped").

skills/transloco-migration/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+
+## 3. Migrating from Angular's built-in i18n (`@angular/localize`)
+
+```bash
+npm install -D @jsverse/transloco-schematics
+ng g @jsverse/transloco-schematics:ng-migrate
+```
+
+- Extracts `i18n`/`i18n-<attr>` strings from HTML into a translations JSON file, converting the string to a kebab-case key (`"My sample string"` → `"my-sample-string"`).
+- `i18n="context|comment"` becomes a `.comment`-suffixed companion key (see transloco-core's comments-for-translators convention); `@@myId` custom IDs become the object's nested key.
```
shaharkazaz: Reversed. The context (the part before `|`) becomes the nested parent object. `@@id` replaces the key itself, after `dasherize` (so `myId` becomes `my-id`), and `.comment` keys are created even without a context (`libs/transloco-schematics/src/ng-migrate/ng-migrate.ts:71-104`). There's also no "comments-for-translators convention" in transloco-core to link to.

skills/transloco-migration/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+- Replaces `i18n`/`i18n-<attr>` directives with the `transloco` pipe.
+
+## 4. Upgrading Transloco itself (v8 → v9)
+
+v9 is currently alpha under the npm `next` tag (`@jsverse/transloco@next`). Confirm the user actually wants to move to alpha/v9 before applying this — otherwise stay on the current v8.x stable line.
+
+```bash
+ng update @jsverse/transloco --next   # or: nx migrate @jsverse/transloco@next && nx migrate --run-migrations
```
shaharkazaz: `nx migrate @jsverse/transloco@next` won't run the v9 migration on the alpha. The migration is versioned `9.0.0` (`libs/transloco/migrations/migration.json:4`), and Nx keeps the prerelease tag when comparing (`nx/src/command-line/migrate/migrate.js`), so `9.0.0` isn't `<= 9.0.0-alpha.1` and it's skipped. `ng update --next` does run it, because the Angular CLI strips the prerelease. I'd drop the nx alternative until 9.0.0 stable (our own guide has the same issue).

skills/transloco-migration/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+- Adding `provideGlobalTranslateFn()` to root providers wherever the standalone `translate()`/`translateObject()` functions are imported from `@jsverse/transloco` (detection is import-anchored, so a locally defined `translate` helper isn't mistaken for it). **The schematic has no SSR/MFE detection** — it adds the provider to every matching application project regardless. For SSR or multi-instance/micro-frontend apps, manually remove the generated provider afterwards and inject `TranslocoService` directly there instead.
+- Declaring/reporting new peer-dependency floors: `@angular/core >=20`, `rxjs ^6.5.3 || ^7.4.0` for Transloco/Locale/Messageformat; `@angular/core >=20` for Persist Lang/Persist Translations/Preload Langs; `@angular/compiler >=20` + `typescript >=5.8` for Keys Manager. `ng update` refuses to proceed if the workspace doesn't satisfy the Angular floor.
```
shaharkazaz: These aren't automated by the migration. The floors are just peer ranges in each `package.json`, and the migration only reports the Node floor (`report-version-floors.ts`). The migration guide's table marks them "Declared", and the keys-manager move "Manual". Please move this and L77 out of "The migration automates".

skills/transloco-migration/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+- Reporting (not enforcing) the new **Node.js >=22** requirement for the CLI packages (`transloco-keys-manager`, `transloco-optimize`, `transloco-scoped-libs`, `transloco-utils`, `transloco-validator`).
+- `@jsverse/transloco-keys-manager` moves into the main `jsverse/transloco` monorepo and jumps straight to the shared `9.0.0` version line — its CLI/config are unchanged.
+
+Full detail: see `BREAKING_CHANGES.md` and `docs/v9/migration-guides/migrate-to-v9.md` in this repository — treat those (and the actual source under `libs/*/src`) as authoritative over any other summary, including this file, if they ever disagree.
```
shaharkazaz: `docs/v9/...` doesn't exist in this repository. Master has no `docs/` folder; the docs live on the `gitbook-docs` branch and are published at jsverse.gitbook.io. Link the live page instead.

skills/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+# A single skill
+npx skills add jsverse/transloco --skill transloco-core
+```
+
+## Maintaining
+
+When Transloco's public API changes:
+
+1. Update the relevant `docs/v*` page(s) as usual.
```
shaharkazaz: Contributors can't follow this rule: there's no `docs/` on master (docs live on the `gitbook-docs` branch), so a docs change and a skills change can't land in the same PR. The same applies to the `docs/` mention on L3.

PR #1017, opened 2026-09-13
chore: migrate to inject() function

medbenmakhlouf:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Code style update (formatting, local variables)
- [ ] Build related changes
- [ ] CI related changes

Several classes across `transloco`, `transloco-messageformat`, and `transloco-preload-langs` still used constructor-parameter dependency injection (`@Inject`/`@Optional`), which Angular now recommends migrating away from in favor of the `inject()` function (see https://angular.dev/reference/migrations/inject-function). The `@angular-eslint/prefer-inject` lint rule, which enforces this convention, was also explicitly disabled for these projects.

Issue Number: N/A

## What is the new behavior?

- Ran the official `@angular/core:inject-migration` schematic to convert constructor-parameter injection to the `inject()` function across `TranslocoService`, `TranslocoPipe`, `DefaultFallbackStrategy`, `TestingLoader`, `MessageFormatTranspiler`, and `TranslocoPreloadLangsService`.
- Fixed fallout from the migration:
  - `pipe.spec.ts` previously instantiated `TranslocoPipe` directly via `new TranslocoPipe(...)`, bypassing Angular DI. Updated it to use `Injector.create` + `runInInjectionContext` so field-initializer `inject()` calls have a valid injection context.
  - `TranslocoService.loader` properly handles the `{ optional: true }` case (which can return `null`) using `??` with a `DefaultLoader` fallback instead of an unsafe cast.
  - Renamed a field in `MessageFormatTranspiler` (`config` → `mfConfigInput`) to avoid silently shadowing the parent `DefaultTranspiler.config` field, which broke interpolation matching in message formatting.
  - Consolidated remaining constructor-only injections (`TranslocoService`, `MessageFormatTranspiler`, `TranslocoPreloadLangsService`) and derived fields (`TranslocoPipe.scopeResolver`) into field initializers for consistency, keeping constructors reserved for classes with genuine side effects/initialization logic (`TranslocoPersistLangService`, `TranslocoLocaleService`).
- Re-enabled `@angular-eslint/prefer-inject` (previously `'off'`) in the root and per-project ESLint configs now that the migration is complete, so future PRs are enforced by lint.

### Follow-up from review

Addressing [@ShaharKazaz](https://github.com/ShaharKazaz)'s feedback to introduce named `injectX()` helpers for injection tokens (scoped to the core `transloco` lib, since it's the main consumer of these tokens):

- Added `injectLoader()` / `injectTranslocoConfig()` for `TRANSLOCO_LOADER` / `TRANSLOCO_CONFIG`, the two tokens with real `inject(TOKEN, { optional: true }) ?? default` fallback logic worth encapsulating in one place.
  - `TRANSLOCO_CONFIG` already has a DI `factory: () => defaultConfig`, so its `?? defaultConfig` fallback was dead code — `injectTranslocoConfig()` now just does `inject(TRANSLOCO_CONFIG)`.
- Added the same helper pattern for the remaining tokens for consistency: `injectTranspiler()`, `injectMissingHandler()`, `injectInterceptor()`, `injectFallbackStrategy()`, `injectTranslocoLang()`, `injectLoadingTemplate()`, `injectTranslocoScope()`.
- Fixed a real bug found along the way: `TRANSLOCO_SCOPE` was typed as `InjectionToken<TranslocoScope>`, but consumers actually need `OrArray<TranslocoScope>` and were using an explicit `inject<...>()` generic to compensate. Corrected the token's own type instead.
- Simplified `TranslocoService`'s manually tracked `destroyed` boolean to use `DestroyRef#destroyed` directly (available since Angular 20, matching this lib's peer dependency).
- Marked injected fields `readonly` where they're never reassigned.

All affected projects (`transloco`, `transloco-messageformat`, `transloco-preload-langs`) pass `lint`, `test`, and `build`, and the full workspace `lint`/`test`/`build`/`format:check` runs are clean.

- [ ] Yes
- [x] No

## Other information

Follows Angular's official migration guide: https://angular.dev/reference/migrations/inject-function

## Summary by CodeRabbit

- **Refactor**
  - Updated Angular dependency injection across translation services, pipes, fallback handling, locale features, persistence, testing utilities, and message formatting to use the modern `inject()` API.
  - Simplified constructors while preserving optional dependencies, configuration defaults, and existing translation behavior.
  - Added consistent injection support for loaders, transpilers, handlers, interceptors, scopes, languages, and loading templates.

- **Tests**
  - Updated pipe tests to use Angular’s injection context and configured providers.

- **Chores**
  - Re-enabled the Angular ESLint `prefer-inject` rule across affected packages.


libs/transloco/src/lib/transloco.service.ts
```diff
@@ -150,9 +143,23 @@ export class TranslationLoadError extends Error {
 
 @Injectable({ providedIn: 'root' })
 export class TranslocoService {
+  private translations = new Map<string, Translation>();
+  private readonly loader: TranslocoLoader =
+    inject(TRANSLOCO_LOADER, { optional: true }) ??
```
shaharkazaz: We can instroduce injectLoader for internal usage which does thi OOTB, I would introduce this for all the injection tokens

medbenmakhlouf: I split it up for easier review, so I : 
- Added `injectLoader()` / `injectTranslocoConfig()` for the two tokens with real fallback logic.
- Applied the same `injectX()` pattern consistently to the other tokens.
- Fixed the `TRANSLOCO_SCOPE` type to `OrArray<TranslocoScope>`, removing the need for explicit generics.
- Kept `transloco-preload-langs` / `messageformat` unchanged since the abstraction doesn’t add much value there.

Happy to revisit if you think the pattern should apply everywhere.

shaharkazaz: I think it should as a convention yes

shaharkazaz (approved):

PR #1019, opened 2026-09-13
fix(transloco): unify TranslocoPipe/TranslocoDirective translation resolution and fix lang precedence bugs

medbenmakhlouf:
## PR Checklist

## PR Type

- [x] Bugfix

`TranslocoPipe` and `TranslocoDirective` each duplicated the logic to resolve the
active lang/scope, load the (possibly scoped) translation, and track which lang
was used for the last emission via mutable per-instance state. That duplicated
logic also had two latent bugs:
- The `LangResolver` instance (and its sticky `initialized` flag) was a shared
  field, so a second `resolve()`/`transform()` call with a different inline lang
  on the same pipe/directive instance would silently keep using the previously
  resolved lang.
- The "should we listen to lang changes" check read `providerLang || inlineLang`,
  which is the opposite precedence of the actual lang resolution (`inline` wins
  over `provider`), so an inline `|static` marker could be ignored if a provider
  lang was also set.

Issue Number: N/A

## What is the new behavior?

- Extracted a shared `TranslationResolver` class (`translation-resolver.ts`) used by
  both `TranslocoPipe` and `TranslocoDirective`.
- `resolve()` emits a single `ResolvedTranslation` (`{ lang, translation }`) per
  emission — for array scopes, all resolved scopes are forkJoin'd and collapsed into
  one result, since they always share the same active lang.
- Consumers now read `lang` directly off the emitted value instead of relying on a
  separate mutable `path` field / `resolveLang()` call after subscribing.
- `LangResolver` is now created per `resolve()` call instead of being a shared field,
  so its `initialized` flag stays scoped to that call and correctly picks up a new
  inline lang on subsequent `resolve()`/`transform()` calls.
- Fixed the lang-changes listening check to use `inlineLang || providerLang`,
  matching `LangResolver`'s actual inline-over-provider precedence.
- Added `resolveSignal()`, a signal-based counterpart of `resolve()` for future
  signal-input-driven consumers.
- Added unit tests for `TranslationResolver` (`translation-resolver.spec.ts`,
  including regression tests for the two bugs above) and updated `pipe.spec.ts`
  for the new emission shape.

- [ ] Yes
- [x] No

## Other information

Public API of `TranslocoPipe`/`TranslocoDirective` is unchanged; the bugfixes are
internal to the newly-extracted `TranslationResolver`.

## Summary by CodeRabbit

- **Improvements**
  - Improved translation language and scope resolution in directives and pipes.
  - Improved handling of multiple translation scopes, inline language settings, and static translations.
  - Translation results now consistently retain resolved language information.
  - Added reactive signal-based translation updates when languages or parameters change.
  - Improved consistency across repeated translation resolutions, including changes to languages and parameters.


libs/transloco/src/lib/translation-resolver.ts
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * Shared by `TranslocoDirective` and `TranslocoPipe`: resolves the active lang
+ * for the given inline/provider lang + scope, and loads the (possibly scoped)
+ * translation for it, respecting `reRenderOnLangChange`/the `|static` suffix.
+ */
+export class TranslationResolver {
+  private scopeResolver: ScopeResolver;
+
+  constructor(private service: TranslocoService) {
```
shaharkazaz: Why pass in the constructor and not make it injectable?

libs/transloco/src/lib/translation-resolver.ts
```diff
@@ -0,0 +1,122 @@
...
+  resolveSignal(
+    paramsFn: () => ResolveTranslationParams,
+  ): Signal<ResolvedTranslation | undefined> {
+    const params$ = toObservable(computed(paramsFn));
+    return toSignal(params$.pipe(switchMap((params) => this.resolve(params))), {
+      initialValue: undefined,
+    });
+  }
+
+  private resolveScope(
```
shaharkazaz:
```suggestion
  #resolveScope(
```

shaharkazaz (commented):
🔥

shaharkazaz (approved):

PR #1020, opened 2026-09-16
chore: migrate to pnpm 11 and nx 23

shaharkazaz:
## What

Moves the repo from npm to **pnpm 11** and upgrades **Nx 22.7.8 → 23.2.1**. The two have to move together (see below).

- **pnpm.** `packageManager: pnpm@11.27.0`. `pnpm-lock.yaml` started as a `pnpm import` of `package-lock.json`, so versions outside the Nx 23 subtree stay as npm had them. The Nx 23 changes were resolved under pnpm 11's default one-day `minimumReleaseAge`, which is kept.
- **`pnpm-workspace.yaml`** lists only the root package, matching Nx's single-version policy. Listing `libs/*` would make pnpm install each lib's own dependency ranges. Its `allowBuilds` approves the build scripts npm used to run (swc, esbuild, nx, lmdb, msgpackr-extract, @parcel/watcher, unrs-resolver) and blocks `less`, whose postinstall downloads Playwright browsers.
- **Undeclared dependencies.** Declares four packages the code imports but only had through npm's hoisting: `@angular-devkit/core`, `@angular-devkit/schematics` and `@schematics/angular` (via `@angular/cli`), plus `domhandler` (via `cheerio`, type-only). They are pinned to versions already in the lockfile.
- **Nx 23**, and only Nx. `nx migrate` also proposed Angular 22 and TypeScript 6, which are left for their own PR. Its TS 6 tsconfig migrations ran even though TypeScript stays on 5.9, so they were reverted. Every other migration was a no-op here: the eslint configs are already flat, and nothing uses `ban-types`. Nx 23 no longer installs the webpack/rspack stack for `@nx/angular`, which is why the lockfile shrinks.
- **Commands.** CI (`step-setup`, `ci.yml`, `release.yml`, `deploy-demo-app.yml`), the husky hooks, `pkg-pr-new` (`pnpm dlx`), CLAUDE.md, CONTRIBUTING.md and the repo's agent files all use pnpm now.
- **Formatting.** `pnpm-lock.yaml` is added to `.prettierignore`, since `ci:format` checks the whole repo.
- **Left on npm on purpose.** `verify-keys-manager-compat.mts` simulates a consumer install outside the workspace, so it keeps using npm. It now has a comment saying so.

## Why Nx 23 is required, and why pnpm 11 rather than 12

- **Nx 22 can't publish with pnpm 11+.** Nx 22's `nx release publish` passes `--@scope:registry=…`, which pnpm 11+ rejects. Nx 23 passes `--config.@scope:registry` for pnpm ≥ 11 ([nrwl/nx#36860](https://github.com/nrwl/nx/issues/36860); compared in `src/utils/package-manager.js` of `nx@22.7.12` and `nx@23.2.1`).
- **pnpm 12 isn't ready for this release setup.** It is three weeks old and has a rewritten `publish`. pnpm itself hasn't tested that publish end to end with OIDC ([pnpm/pnpm#12738](https://github.com/pnpm/pnpm/issues/12738)) or with provenance signing ([#12739](https://github.com/pnpm/pnpm/issues/12739)), and `release.yml` relies on both. Nx also still has an open pnpm 12 bug ([nrwl/nx#37032](https://github.com/nrwl/nx/issues/37032)).
- **pnpm 11 has a working publish path.** Its native OIDC publish is fixed, and users confirm it works ([pnpm/pnpm#11513](https://github.com/pnpm/pnpm/issues/11513), [#11566](https://github.com/pnpm/pnpm/issues/11566)).

## Release path (OIDC trusted publishing)

- **Publish command.** `nx release publish` → `pnpm publish` (native in pnpm 11). `registry-url` is still left unset in `step-setup`.
- **Action version.** `pnpm/action-setup@v6` resolves to v6.0.10, past the v6.0.6 fix that native OIDC publishing needs.
- **Tarballs.** I packed all 12 release packages with both `npm pack` and pnpm 11 `pnpm pack`. The file lists are identical, LICENSE included. The two `package.json` files I diffed (`@jsverse/transloco` and `@jsverse/transloco-schematics`) are identical too.
- **Rehearsal.** `nx release publish` (Nx 23 + pnpm 11) published all 12 packages under a throwaway version to a local verdaccio (`nx run transloco-nx:local-registry`). This covers the registry flag that broke under Nx 22, but not the OIDC login.

> [!WARNING]
> Nothing in CI can prove the OIDC publish works. `release.yml`'s `--dry-run` doesn't authenticate, so the first real release is the first real test of that path.

## Verified locally

These ran in a copy outside the npm checkout, so no parent `node_modules` could supply a missing package: `pnpm install --frozen-lockfile`, `ci:build`, `ci:test`, `ci:lint`, `ci:e2e`, `ci:format` (the only files flagged were gitignored keys-manager test output), `verify:license`, the keys-manager compat check (ng22 / ts6.0) and `actionlint`. The pre-commit and commit-msg hooks ran through pnpm on these commits.

## Contributor note

Existing clones need `rm -rf node_modules && pnpm install`. pnpm will switch itself to 11.27.0 from `packageManager`.


package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.0.0",
   "license": "MIT",
   "homepage": "https://jsverse.gitbook.io/transloco/",
+  "packageManager": "pnpm@11.27.0",
```
medbenmakhlouf: Nit: an accidental `npm install` still works and brings back `package-lock.json`. Moving the pin to `devEngines` makes npm refuse (`EBADDEVENGINES`). pnpm and `pnpm/setup` both read it and still switch to 11.27.0. Replace `packageManager` rather than adding next to it, because pnpm warns when both are set and ignores `packageManager`.

```suggestion
  "devEngines": {
    "packageManager": {
      "name": "pnpm",
      "version": "11.27.0",
      "onFail": "error"
    }
  },
```


`pnpm-lock.yaml` must be updated too, because pnpm 11 pins the pnpm binary in it. Without that, `pnpm install --frozen-lockfile` fails with:

```
Cannot update packageManagerDependencies with "frozen-lockfile" because the lockfile is not up to date
```

medbenmakhlouf (dismissed):

PR #1021, opened 2026-09-18
feat(transloco): add TranslocoTitleStrategy to translate and reactively update router titles

### LINKED ISSUE #839: Feature(scope): Route Title Resolver

- [x] I have searched the existing issues

### Which Transloco package(s) will this feature affect?

Don't know / other

### Is your feature request related to a problem? Please describe

Hi everyone,

I'm looking for a way to translate every tab title without duplicating the code. To solve this, I would like to create a resolver such as:
```ts
export const titleTranslationResolver: ResolveFn<string> = (route, state) => {
  console.log("titleTranslationResolver")
  if (!route.data["titleKey"]) {
    return Promise.resolve('Default title');
  }

  const transloco = inject(TranslocoService);
  return Promise.resolve(transloco.translate(route.data["titleKey"]));
};
```

and use it in my router:
```ts
export const routes: Routes = [
  {
    path: 'participants',
    title: titleTranslationResolver,
    pathMatch: 'full',
    loadComponent: () => import('@presentation/pages/participants-page/participants-page.component').then(m => m.ParticipantsPageComponent),
    canActivate: [ConnectedGuard],
    data: { titleKey: 'participant_list' }
  },
  { path: '',   redirectTo: 'participants', pathMatch: 'full' },
  {
    path: '**',
    title: titleTranslationResolver,
    loadComponent: () => import('@presentation/pages/not-found-page/not-found-page.component').then(m => m.NotFoundPageComponent)
  }
];
```

The issue with this solution is that Transloco is initialized after my routes, even with this code:
```ts
export const appConfig: ApplicationConfig = {
  providers: [
    provideHttpClient(),
    provideTransloco({
      config: {
        availableLangs: ['en'],
        defaultLang: 'en',
        // Remove this option if your application doesn't support changing language in runtime.
        reRenderOnLangChange: true,
        prodMode: !isDevMode(),
      },
      loader: TranslocoHttpLoader
    }),
    provideRouter(routes),
  ]
};
```

Image

### Describe the solution you'd like

Maybe I did something wrong, but here are two potential solutions:
- Use a native Transloco resolver for translating tabs.
- Allow the TranslocoHttpLoader to initialize before the router.

### Describe alternatives you've considered

```markdown
For now I translate in every component "page" in the ngOnInit method
```

### Additional context

_No response_

### I would like to make a pull request for this feature

No

medbenmakhlouf:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Angular's Router resolves a route's `title` (whether a plain string or a
`ResolveFn<string>`) once per navigation, via `DefaultTitleStrategy`. Today,
translating that title requires a custom `ResolveFn` that calls
`TranslocoService.translate()` directly (e.g. `() => inject(TranslocoService).translate(key)`).
Because the resolved title is only computed at navigation time, the document
title does **not** update when the user changes the active language — it's
stuck with the language that was active during the last navigation until a
hard refresh or another navigation occurs.

Issue Number: Closes #839

## What is the new behavior?

Adds `TranslocoTitleStrategy`, a `TitleStrategy` (from `@angular/router`) that:

- Treats the route's resolved `title` as a Transloco translation key (rather
  than literal text) and translates it before setting it via `Title.setTitle()`.
- Subscribes once to `TranslocoService.langChanges$` and re-translates/re-applies
  the cached title whenever the active language changes — no new navigation or
  hard refresh required.
- Also re-applies the title whenever a translation successfully loads
  (`TranslocoService.events$`, `translationLoadSuccess`), fixing the case
  (raised by [@windmichael](https://github.com/jsverse/transloco/issues/839#issuecomment-2888468187)
  and [@MarcScheib](https://github.com/jsverse/transloco/issues/839#issuecomment-3140591905)
  on #839) where a scoped/lazy-loaded title key isn't available yet on the
  *first* navigation to its route.
- Cleans up the subscription in `ngOnDestroy`.

Also adds `provideTranslocoTitleStrategy()` to register it as the Router's
`TitleStrategy` with one line in `app.config.ts`.

Usage:

```ts
// app.config.ts
import { provideTranslocoTitleStrategy } from '@jsverse/transloco/router';

providers: [provideTranslocoTitleStrategy()];

// app.routes.ts
export const routes: Routes = [
  {
    path: 'design-system',
    loadComponent: () => import('./design-system.page'),
    title: 'app.menu.design_system', // a translation key, not literal text
  },
];
```

`TranslocoTitleStrategy`/`provideTranslocoTitleStrategy` live in a new
secondary entry point, `@jsverse/transloco/router`, so importing
`@jsverse/transloco` never statically pulls in `@angular/router`/
`@angular/platform-browser` — those are declared as **optional** peer
dependencies in `libs/transloco/package.json`, and only consumers who
explicitly import from `@jsverse/transloco/router` need them installed.

### `transloco-keys-manager` integration

`transloco-keys-manager` now auto-extracts these route `title` keys, **no
`marker`/`_()` wrapping needed** for the common case above:

- It detects `provideTranslocoTitleStrategy()` usage anywhere in the project
  (it's typically called once, in `app.config.ts`, far from the route files).
- It then treats a plain-string `title` as a translation key **only** on
  object literals that look like a `Route` (i.e. have both a route locator —
  `path`/`matcher` — and a route shaper — `component`/`loadComponent`/
  `loadChildren`/`children`/`redirectTo`), so it won't misfire on unrelated
  `{ path, title }`-shaped data (e.g. a breadcrumb config).
- A `title` that's a `ResolveFn` (a function) or an empty string is left
  alone.
- A scoped title is a plain string prefixed with the scope alias
  (`title: 'admin.title'`), the same as `{{ 'admin.title' | transloco }}`.
  A known scope-alias prefix (from `provideTranslocoScope(...)`, wherever it's
  declared, e.g. the route's or component's `providers`) is extracted into that
  scope's file; any other prefix (e.g. `app.menu.*`) stays global. The
  `marker(key, undefined, scope)` form is not supported for titles, since
  `marker()` returns its first argument at runtime and the scope would be lost.
- At runtime the strategy doesn't load scopes itself: a scoped title resolves
  once the page uses its scope (pipe/directive/`translateSignal`), and shows
  the raw key until then.
- We considered following the actual router graph (starting from
  `provideRouter(routes)` in `app.config.ts`, through `loadChildren`/
  `children`) instead of this shape-based heuristic. It would be more
  precise, but requires a module-resolution layer keys-manager doesn't have
  today (resolving import specifiers, path aliases, barrel re-exports), with
  real false-negative risk if resolution misses a file — worse than the
  heuristic's already-closed false-positive risk. Skipped for now.

- [ ] Yes
- [x] No

## Other information

📖 GitBook docs for this feature (under the `transloco` plugin/router section)
still need to be written in a follow-up PR against the `gitbook-docs` branch.

## Summary by CodeRabbit

* **New Features**
  * Added Angular router integration for translating document titles from route configuration.
  * Document titles update when the active language changes or lazy-loaded translations become available.
  * Added a provider and dedicated package entry point for translated route titles.
  * Translation key extraction now supports eligible static and scoped route titles.

* **Bug Fixes**
  * Improved route-title detection to exclude dynamic, empty, nested, non-route, and unconfigured titles.

* **Compatibility**
  * Added support for Angular 20 or later router and browser platform integrations.


libs/transloco/router/src/title-strategy.ts
shaharkazaz, line 1: Rename to title-strategy.ts and move under a folder called router, it will align with some renames I'm doing

medbenmakhlouf: done

libs/transloco-keys-manager/src/lib/keys-builder/typescript/index.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+      function provideTranslocoTitleStrategy() {
+        return {};
+      }
+
+      export const appConfig = {
+        providers: [provideTranslocoTitleStrategy()],
+      };
+    `);
+
+    expect(isTitleStrategyProviderCalled(ast)).toBe(false);
```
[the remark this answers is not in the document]
medbenmakhlouf: done

libs/transloco/router/src/title-strategy.ts
```diff
@@ -0,0 +1,115 @@
...
+  override updateTitle(snapshot: RouterStateSnapshot): void {
+    this.titleKey = this.buildTitle(snapshot);
+
+    if (this.titleKey !== undefined) {
+      this.applyTitle(this.titleKey);
+    }
+  }
+
+  private applyTitle(key: string): void {
+    this.title.setTitle(this.transloco.translate(key));
```
shaharkazaz: **Scoped titles never resolve (the documented `marker(key, undefined, scope)` recipe is broken at runtime).** `marker()` simply returns its first argument, so `title: _('title', undefined, 'admin')` makes the route title the string `'title'`. `applyTitle` then calls `this.transloco.translate('title')` with no scope, which looks up the global key `title`. Once the `admin` scope loads, its keys are stored in the active lang as `admin.title`, so the lookup misses, and the tab shows the raw key `title` (or whatever the missing handler returns), even after the `translationLoadSuccess` re-apply. The recipe in the JSDoc, the PR description, and the `route-title` fixture only works for extraction and never works at runtime. The recipe that does work is a fully-qualified plain string (`title: 'admin.title'`), and it hits the extractor issue I flagged in `route-title.extractor.ts`.

libs/transloco-keys-manager/src/lib/keys-builder/typescript/route-title.extractor.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    if (!hasLocator || !hasShaper || !titleProperty) continue;
+
+    const { initializer } = titleProperty;
+    const isPlainStringTitle =
+      ts.isStringLiteral(initializer) ||
+      ts.isNoSubstitutionTemplateLiteral(initializer);
+
+    if (isPlainStringTitle && initializer.text.length > 0) {
+      result.push({ key: initializer.text, lang: '', params: [] });
```
shaharkazaz: **Scope-prefixed titles are extracted into the global file instead of the scope file.** The extractor always pushes `lang: ''`, so `resolveAliasAndKeyFromService` returns scope `null` and `title: 'admin.title'` gets written as the key `admin.title` in the **global** translation file. At runtime the strategy's `translate('admin.title')` resolves against the loaded `admin` scope (`admin/en.json` → `title`), so the scope file never gets the key and the global file picks up a dead one. Template extraction already handles this: `resolveAliasAndKey(key, scopes)` in `utils/resolvers.utils.ts` splits off a known scope-alias prefix (`{{ 'admin.title' | transloco }}`). Using it here would handle scoped titles and remove the need for the `marker(..., scope)` workaround, which doesn't work at runtime anyway (see my comment on `title-strategy.ts`).

libs/transloco-keys-manager/src/lib/keys-builder/typescript/index.ts
```diff
@@ -11,29 +12,261 @@ import { readFile } from '../../utils/file.utils';
...
+ *   (not called) doesn't count;
+ * - an unrelated local function/variable that happens to share the name
+ *   `provideTranslocoTitleStrategy` (but isn't imported from the package)
+ *   doesn't count either;
+ * - both `import { provideTranslocoTitleStrategy as provideTitle } from '...'`
+ *   and `import * as router from '...'; router.provideTranslocoTitleStrategy()`
+ *   are correctly recognized as the real provider being used.
+ */
+function detectTitleStrategyProvider(config: Config): boolean {
+  return resolveFileList(config, 'ts').some((file) => {
```
shaharkazaz: **Route titles aren't extracted in the webpack plugin's watch mode.** The provider check runs on `resolveFileList(config, 'ts')`, which returns `config.files` when that is set. `webpack-plugin.ts` calls `extractTSKeys({ ...this.config, files: keysExtractions.ts })` with only the modified files. So in watch mode, after editing `app.routes.ts` to add `title: 'app.menu.new'`, the detection only looks at `app.routes.ts`, finds no `provideTranslocoTitleStrategy()` call, returns false, and the new title key is silently not extracted (unless `app.config.ts` happens to be in the same change batch). The detection should always scan the full `input` globs, ignoring `files`, or its result should be computed once and cached on the plugin/config.

libs/transloco/router/src/title-strategy.ts
medbenmakhlouf: Done

libs/transloco-keys-manager/src/lib/keys-builder/typescript/route-title.extractor.ts
medbenmakhlouf: Done

libs/transloco-keys-manager/src/lib/keys-builder/typescript/index.ts
```diff
@@ -11,29 +12,262 @@ import { readFile } from '../../utils/file.utils';
...
+
+  return extractKeys(config, 'ts', (extractorConfig) =>
+    TSExtractor(extractorConfig, hasTitleStrategyProvider),
+  );
 }
 
 const translocoImport = /@(jsverse|ngneat)\/transloco/;
 const translocoKeysManagerImport = /@(jsverse|ngneat)\/transloco-keys-manager/;
-function TSExtractor(config: ExtractorConfig): ScopeMap {
+const translocoRouterImport = /@(jsverse|ngneat)\/transloco\/router/;
```
shaharkazaz: Why include all this logic here? it bloated the index with 243 LOC

shaharkazaz: drop ngneat as it's not published in that scope, it was only for backward compatibility for existing features

libs/transloco-keys-manager/src/lib/keys-builder/typescript/route-title.extractor.spec.ts
shaharkazaz, line 1: How do you verify that we actually extract title from a route object? if you add a test case for an object with the key title, does it extract it?

medbenmakhlouf: I scanned the properties that characterize the Angular Router object — see lines 10, 79, and 80 where I checked this. 
```ts
const LOCATOR_PROPERTIES = new Set(['path', 'matcher']);
const SHAPER_PROPERTIES = new Set([
  'component',
  'loadComponent',
  'loadChildren',
  'children',
  'redirectTo',
]);
```

I also covered this case in `libs/transloco-keys-manager/src/lib/tests/buildTranslationFiles/ts-extraction/route-title/src/app.routes.ts`, line 70.

libs/transloco-keys-manager/src/lib/keys-builder/typescript/index.ts
medbenmakhlouf: good catch! this is done

medbenmakhlouf: should I keep ngneat for `translocoImport ` and `translocoKeysManagerImport ` ?

shaharkazaz: Let's keep this PR focused

libs/transloco-keys-manager/src/lib/keys-builder/typescript/route-title.extractor.spec.ts
shaharkazaz, line 1: Apply to all spec files, use GWT pattern

libs/transloco-keys-manager/src/lib/keys-builder/typescript/title-strategy-provider.detector.ts
```diff
@@ -0,0 +1,224 @@
...
+ *   (not called) doesn't count;
+ * - an unrelated local function/variable that happens to share the name
+ *   `provideTranslocoTitleStrategy` (but isn't imported from the package)
+ *   doesn't count either;
+ * - both `import { provideTranslocoTitleStrategy as provideTitle } from '...'`
+ *   and `import * as router from '...'; router.provideTranslocoTitleStrategy()`
+ *   are correctly recognized as the real provider being used.
+ */
+export function detectTitleStrategyProvider(config: Config): boolean {
+  return resolveFileList(config, 'ts').some((file) => {
```
shaharkazaz: I wonder if running on all the files to begin with makes sense and not collect the values as you go and just ignore them in the end if it wasn't initiated

libs/transloco-keys-manager/src/lib/tests/buildTranslationFiles/ts-extraction/route-title-no-provider/src/app.config.ts
```diff
@@ -0,0 +1,11 @@
+import { ApplicationConfig } from '@angular/core';
+// The provider is only imported and mentioned in a comment below - it's
+// never actually called, so `provideTranslocoTitleStrategy()`-detection must
+// treat this project as NOT using the title strategy.
+import { provideTranslocoTitleStrategy } from '@jsverse/transloco/router';
```
shaharkazaz: what about type imports? do we handle those?

medbenmakhlouf: With the simplified detection any import of the provider counts, type-only included.

libs/transloco-keys-manager/src/lib/keys-builder/typescript/title-strategy-provider.detector.ts
medbenmakhlouf: Done, detection runs in the same pass and keys are only applied if the provider import is found.

libs/transloco-keys-manager/src/lib/keys-builder/typescript/route-title.extractor.spec.ts
medbenmakhlouf: Done

libs/transloco/router/src/title-strategy.ts
```diff
@@ -1,9 +1,52 @@
...
+  whenMissing?: 'wait' | 'key';
+
+  /**
+   * Applied to the translated title right before `Title.setTitle()`, e.g.
+   * to append an app name: `(title) => \`${title} - MyApp\``.
+   */
+  format?: (translatedTitle: string) => string;
+}
+
+export const TRANSLOCO_TITLE_STRATEGY_CONFIG =
```
shaharkazaz: Why do we export this token if the config as part of the provide call?

medbenmakhlouf: It was unintentional, just something I did instinctively But you’re right, there’s no need to export it.

shaharkazaz (approved):

PR #1024, opened 2026-09-19
feat(keys-manager)!: remove the webpack plugin

shaharkazaz:
## PR Checklist

- [ ] The commit message follows our guidelines: https://github.com/jsverse/transloco/blob/master/CONTRIBUTING.md#commit

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe: removes the keys-manager webpack plugin

`@jsverse/transloco-keys-manager` ships `TranslocoExtractKeysWebpackPlugin`, a webpack `watchRun` plugin that extracts keys on rebuild. It only works with webpack-based builders (via `ngx-build-plus`), not with Angular's default esbuild/Vite builder, and it runs its own incremental extraction path that has to be kept in sync with the CLI's.

The `keys-manager` schematic asks for a `strategy` (`CLI` / `Webpack Plugin` / `Both`) and, for the plugin options, installs `ngx-build-plus`, writes `webpack-dev.config.js` and switches the `serve` builder to `ngx-build-plus:dev-server`.

Issue Number: N/A

## What is the new behavior?

- The `webpack-plugin/` sources (`webpack-plugin.ts`, `generate-keys.ts`) are deleted.
- The package's root entry point is removed. It only re-exported the plugin (`marker` already moved to `@jsverse/transloco-keys-manager/marker` in v9), so `exports` is now `./marker` + `./package.json`, and `main`/`types` are gone. The `transloco-keys-manager` CLI (`bin`) is unchanged.
- The `keys-manager` schematic drops the `strategy` option: it installs only `@jsverse/transloco-keys-manager` and adds the `i18n:extract` / `i18n:find` scripts.
- README feature bullet and the `Webpack` package keyword are removed.

- [x] Yes
- [ ] No

Apps using `TranslocoExtractKeysWebpackPlugin` (typically a `webpack-dev.config.js` passed via `ngx-build-plus`) need to remove it and run `transloco-keys-manager extract` instead, e.g. as a script alongside `ng serve`. Anything importing the bare `@jsverse/transloco-keys-manager` specifier will no longer resolve; the only remaining runtime export is `marker` from `/marker`. The v9 `ng update` migration still moves `marker` to `/marker`, and now warns with the file list for anything left pointing at the package root (e.g. a plugin co-import or a `webpack-dev.config.js` `require`). It doesn't delete those references, because the plugin has no automatic replacement.

## Other information

- No existing tests covered the plugin or the `keys-manager` schematic, so none were removed or updated. `migration-v9.spec.ts` gains cases for the root-reference warning (plugin co-import, `.js` webpack config, and no warning for a lone `marker` import).
- Deliberately not touched here, to avoid conflicting with #1021: the incremental `files`-based extraction path in `keys-builder/` that the plugin drove. It can be revisited once #1021 lands.
- Follow-up: the GitBook docs (`gitbook-docs` branch) still document the webpack plugin.

## Summary by CodeRabbit

* **Changes**
  * Removed Webpack-based translation key extraction support from the Keys Manager.
  * Simplified Keys Manager setup to use the CLI extraction workflow.
  * Removed the Webpack strategy option and related setup configuration.
  * Updated package documentation and metadata to reflect the supported CLI workflow.
  * Improved migration handling for marker imports across JavaScript and TypeScript module formats, including type-only and aliased imports.
  * Added migration guidance for removed Webpack plugin references and configurations.


medbenmakhlouf:
Nice cleanup! just one thing, The v9 `marker-import` migration still leaves every name except marker on the root import, which no longer exists. So `import { TranslocoExtractKeysWebpackPlugin, marker } from '@jsverse/transloco-keys-manager'` ends up as a broken import after ng update. The migration should remove that import or warn about it.

